PR #1, opened 2023-03-27
mvp

joelrebel:
This PR includes a functional Condition orchestrator API service, API Client, Orchestrator services.

The conditionorc currently implements two main functionalities,

- Provide a CRUD API to set [conditions](https://github.com/metal-toolbox/architecture/blob/main/firmware-install-service.md#conditions) on servers in Serverservice.
- Listen for Server create events from Serverservice and set the` InventoryOutofband` condition for Alloy to reconcile.

To understand conditionorc and its role, check out https://github.com/metal-toolbox/architecture/blob/main/firmware-install-service.md#condition-orchestrator---conditionorc

For information on the events conditionorc sends and receives, check out https://github.com/metal-toolbox/architecture/blob/f5b27686e634fed5b0215379f58418728e42227a/messaging.md#configuration

To test the Conditionorc along with Alloy, check out the https://github.com/metal-toolbox/sandbox


cmd/orchestrator.go
```diff
@@ -0,0 +1,60 @@
...
+		}
+
+		orc := orchestrator.New(options...)
+		go func() {
+			orc.Run(cmd.Context())
+		}()
+
+		// sit around for term signal
+		<-app.TermCh
+		app.Logger.Info("got TERM signal, exiting")
```
mmlb, line 52: should `orc` be terminated cleanly after getting the signal?

cmd/server.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/metal-toolbox/conditionorc/internal/app"
+	"github.com/metal-toolbox/conditionorc/internal/model"
+	"github.com/metal-toolbox/conditionorc/internal/server"
+	"github.com/metal-toolbox/conditionorc/internal/store"
+	"github.com/spf13/cobra"
+	"go.hollow.sh/toolbox/events"
+)
+
+var (
+	shutdownTimeout = time.Second * 10
```
mmlb, line 19: small nit, but imo `10 * time.Second` reads nicer because it matches how we thing of numbers and units aka 10s.

internal/app/app.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"context"
+	"os"
+	"os/signal"
+	"sync"
+	"syscall"
+
+	"github.com/metal-toolbox/conditionorc/internal/model"
+	"github.com/spf13/viper"
+
```
mmlb, line 12: extra new line?

internal/app/app.go
```diff
@@ -0,0 +1,89 @@
...
+// Config holds configuration data when running mctl
+// App holds attributes for the mtl application
+type App struct {
+	// Viper loads configuration parameters.
+	v *viper.Viper
+	// The kind of application - server/orchestrator
+	AppKind model.AppKind
+	// Sync waitgroup to wait for running go routines on termination.
+	SyncWG *sync.WaitGroup
```
mmlb, line 25: this doesn't seem to be used and feels weird to export maybe it can be internal only?

internal/app/app.go
```diff
@@ -0,0 +1,89 @@
...
+	// Flasher configuration.
+	Config *Configuration
+	// TermCh is the channel to terminate the app based on a signal.
+	TermCh chan os.Signal
```
mmlb, line 29: This also feels like a weird export since its not limited to blocking reads for callers. Maybe consider returning a read-only ch in New or some other method?

internal/app/config.go
```diff
@@ -0,0 +1,244 @@
...
+// Server service configuration options
+
+// nolint:gocyclo // parameter validation is cyclomatic
+func (a *App) envVarServerserviceOverrides() error {
+	if a.v.GetString("serverservice.endpoint") != "" {
+		a.Config.ServerserviceOptions.Endpoint = a.v.GetString("serverservice.endpoint")
+	}
+
+	if a.v.GetString("serverservice.facilitycode") != "" {
+		a.Config.ServerserviceOptions.Endpoint = a.v.GetString("serverservice.facilitycode")
```
mmlb, line 181: typo? `.Endpoint` = `...facilitycode`

cmd/orchestrator.go
joelrebel: fixed in [bf26401](https://github.com/metal-toolbox/conditionorc/pull/1/commits/bf26401ad4f1028f03d581c7a48e63581a7f5904)

cmd/server.go
joelrebel: ah right thanks, not sure what I was thinking there :D

internal/app/app.go
joelrebel: fixed in [0335f0e](https://github.com/metal-toolbox/conditionorc/pull/1/commits/0335f0e3da4280dbdc079e734e9f72eb150d9abc)

internal/app/config.go
joelrebel: oh good catch, fixed in [3ab8e8b](https://github.com/metal-toolbox/conditionorc/pull/1/commits/3ab8e8b08bb24946814fe0e3fea2469792695221)

internal/app/app.go
joelrebel: I'd like to keep this exported, for this first MVP, the orchestrator isn't [processing events concurrently](https://github.com/metal-toolbox/conditionorc/blob/mvp/internal/orchestrator/orchestrator.go#L83).
 In the next set of changes the `SyncWG` var will be passed to the `orchestrator` so it processes events in separate routines, and we'll have a `app.SyncWG.Wait()` before orc returns on a TERM signal.

internal/app/app.go
joelrebel: That makes sense, fixed in [83a08ce](https://github.com/metal-toolbox/conditionorc/pull/1/commits/83a08ce04ce5bde99c8c0a625770c2ba819c37c0)

joelrebel:
@mmlb I've addressed your comments, let me know if theres anything else here.

pkg/api/v1/types/types.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/google/uuid"
+	ptypes "github.com/metal-toolbox/conditionorc/pkg/types"
+	"github.com/pkg/errors"
+)
+
+type ServerResponse struct {
+	StatusCode int                 `json:"statusCode,omitempty"`
+	Message    string              `json:"message,omitempty"`
+	Record     *ConditionResponse  `json:"record,omitempty"`
```
DoctorVin, line 15: feels redundant to have a singular and multiple fields that are basically the same thing. Also the difference between `ConditionResponse` and `ConditionsResponse` is hard to discern and a code-maintenance drag. Consider only allowing the multiple-response type and explicitly allowing `len(Records)>=0`.

pkg/types/types.go
```diff
@@ -0,0 +1,140 @@
...
+		if e.Kind == k {
+			return e
+		}
+	}
+
+	return nil
+}
+
+// Condition defines model for Condition.
+type Condition struct {
```
DoctorVin, line 98: As we discussed earlier, we probably want at least two other fields: `ID` (of some kind, preferably string or UUID), `structureVersion` (so we can control upgrades and reading historical data; this probably does not need to be an exported field). Finally we might want to use an exported `Version` or `Revision` field to provide for the locking/updating semantics you reference in the K8s API doc under `concurrency control`.

DoctorVin (commented):
This seems like a reasonable starting point for an MVP, and I'm comfortable merging once we agree on action items to address some of the comments.

cmd/server.go
```diff
@@ -0,0 +1,83 @@
...
+
+		// call server shutdown with timeout
+		ctx, cancel := context.WithTimeout(cmd.Context(), shutdownTimeout)
+		defer cancel()
+		if err := srv.Shutdown(ctx); err != nil {
+			app.Logger.Fatal("server shutdown error:", err)
+		}
+
+		// wait until context is done
+		<-ctx.Done()
```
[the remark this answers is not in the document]
joelrebel: `srv.Shutdown` is being passed a context with a deadline [here](https://github.com/metal-toolbox/conditionorc/pull/1/files/d1a43d342f546a3a0077cd2aee0dc67e7ee89de4#diff-ee2560d627cd77f1563972217a33ac39a4bf56fbf30d7f560fbb4199b9f70197R66), unless you think this may not be enough?

pkg/api/v1/client/client.go
```diff
@@ -0,0 +1,102 @@
...
+func (c *Client) ServerConditionList(ctx context.Context, serverID uuid.UUID, conditionState ptypes.ConditionState) (*v1types.ServerResponse, error) {
+	path := fmt.Sprintf("servers/%s/state/%s", serverID.String(), conditionState)
+
+	return c.get(ctx, path)
+}
+
+func (c *Client) ServerConditionCreate(ctx context.Context, serverID uuid.UUID, conditionKind ptypes.ConditionKind, conditionCreate v1types.ConditionCreate) (*v1types.ServerResponse, error) {
+	path := fmt.Sprintf("servers/%s/condition/%s", serverID.String(), conditionKind)
+
+	return c.post(ctx, path, conditionCreate)
```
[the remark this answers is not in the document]
joelrebel: The [condition create payload](https://github.com/metal-toolbox/conditionorc/blob/d1a43d342f546a3a0077cd2aee0dc67e7ee89de4/pkg/api/v1/types/types.go#L32) has different parameters from the [condition update payload](https://github.com/metal-toolbox/conditionorc/blob/d1a43d342f546a3a0077cd2aee0dc67e7ee89de4/pkg/api/v1/types/types.go#L48).

I'm inclined to keep these separate so they have separate validators in the handlers and its simpler to grok, unless you think it'll be simpler to just have a single PUT handler.

pkg/api/v1/routes/handlers.go
```diff
@@ -0,0 +1,417 @@
...
+	if existing != nil && !ptypes.ConditionStateFinalized(existing.State) {
+		c.JSON(
+			http.StatusBadRequest,
+			&v1types.ServerResponse{Message: "condition present non-finalized state: " + string(existing.State)},
+		)
+
+		return
+	}
+
+	// purge the existing condition
```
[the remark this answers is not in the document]
joelrebel: The PUT server handler payload is limited to [certain fields](https://github.com/metal-toolbox/conditionorc/blob/d1a43d342f546a3a0077cd2aee0dc67e7ee89de4/pkg/api/v1/types/types.go#L48) and so just those fields can be unconditionally clobbered, let me know if I'm missing your point here.

cmd/server.go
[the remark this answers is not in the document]
DoctorVin: Right, I wasn't very clear (and I misread your code 😅 ). Once the `Shutdown` call is complete, there isn't a need to wait on that context to timeout any more, so we can skip lines 73-77 IMO.

pkg/api/v1/routes/handlers.go
```diff
@@ -0,0 +1,417 @@
...
+	if err != nil {
+		c.JSON(
+			http.StatusBadRequest,
+			&v1types.ServerResponse{Message: err.Error()},
+		)
+
+		return
+	}
+
+	if existing == nil {
```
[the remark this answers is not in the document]
joelrebel: good point, fixed in [f8a80d7](https://github.com/metal-toolbox/conditionorc/pull/1/commits/f8a80d79eaf2ee56a85b2ee96493cda01aac763e)

pkg/api/v1/routes/handlers.go
```diff
@@ -0,0 +1,417 @@
...
+	if err != nil && !errors.Is(err, store.ErrConditionNotFound) {
+		c.JSON(
+			http.StatusBadRequest,
+			&v1types.ServerResponse{Message: err.Error()},
+		)
+
+		return
+	}
+
+	if existing != nil && !ptypes.ConditionStateFinalized(existing.State) {
```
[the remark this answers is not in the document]
joelrebel: That makes sense, updated in [7b2d7ba](https://github.com/metal-toolbox/conditionorc/pull/1/commits/7b2d7ba8035faeb33030c6c2a3a80378e9038823)

internal/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,184 @@
...
+		return
+	}
+
+	if err := o.streamBroker.PublishAsyncWithContext(
+		ctx,
+		events.ResourceType(ptypes.ServerResourceType),
+		events.EventType(ptypes.InventoryOutofband),
+		serverID.String(),
+		condition,
+	); err != nil {
```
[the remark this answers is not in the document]
joelrebel: fixed in [937a9a0](https://github.com/metal-toolbox/conditionorc/pull/1/commits/937a9a0e87e10f2378a010f508717ae9e85f228c)

internal/app/app.go
joelrebel: These are good points, for now I've removed the exported  app `SyncWG` field completely and I'll limit its use in the manner suggested above once it becomes necessary for the `orchestrator` component.

pkg/types/types.go
joelrebel: > we probably want at least two other fields: ID (of some kind, preferably string or UUID), `structureVersion`

[d278100](https://github.com/metal-toolbox/conditionorc/pull/1/commits/d27810004464f02ac283fe31b783655d6e054324) adds an ID field, further work on will be followed up in https://github.com/metal-toolbox/conditionorc/issues/3

[e2db03f](https://github.com/metal-toolbox/conditionorc/pull/1/commits/e2db03ffbdaec72525bd5120aacb6115941aeba2) adds the `Version` field - let me know if this is too ambiguous to begin with.

> Finally we might want to use an exported Version or Revision field to provide for the locking/updating semantics you reference in the K8s API doc under concurrency control.

The current [`ResourceVersion`](https://github.com/metal-toolbox/conditionorc/blob/e2db03ffbdaec72525bd5120aacb6115941aeba2/pkg/types/types.go#L133)  covers this ^, unless I'm missing something.

pkg/api/v1/routes/handlers.go
```diff
@@ -0,0 +1,417 @@
...
+				http.StatusInternalServerError,
+				&v1types.ServerResponse{Message: err.Error()},
+			)
+
+			return
+		}
+	}
+
+	// check if any condition with exclusive set is in non-finalized states
+	if errEx := r.exclusiveNonFinalConditionExists(c.Request.Context(), serverID); errEx != nil {
```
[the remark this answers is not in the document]
joelrebel: This was split into its own function since the linter was complaining about the cyclomatic complexity being high for [serverConditionCreate()](https://github.com/metal-toolbox/conditionorc/blob/mvp/pkg/api/v1/routes/handlers.go#L130), unless you mean't something else here - I'm inclined to leave it as is for now..

joelrebel:
@DoctorVin thanks for the feedback.
 I've addressed and resolved most of the suggestions, let me know if you have any strong opinions on the unresolved comments.

pkg/api/v1/routes/handlers.go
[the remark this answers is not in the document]
DoctorVin: I'm OK with leaving this for now. I'll try and refactor a bit, maybe, and see if I can't appease the linter.

pkg/api/v1/client/client.go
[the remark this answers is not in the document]
DoctorVin: Let's keep it like this for now, but be open to refactoring later if we can isolate commonality.

DoctorVin (approved):
LGTM

PR #5, opened 2023-04-14
Vc/simplify condition response

DoctorVin:
#### What does this PR do
Here we simplify the data structures in use by `conditionorc` by removing the single condition-response field in the `ServerResponse` and updating the semantics of that field such that `ConditionsResponse` is used unconditionally, with its slice of `Conditions` allowed to be empty, have one condition, or have multiple.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,7 @@
 # Application binary itself
 /alloy
 /alloy-osx
-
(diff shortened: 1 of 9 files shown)


joelrebel (approved):
LGTM! 🚀

PR #11, opened 2023-04-21
Vc/metrics part1

DoctorVin:
#### What does this PR do
Add prometheus metrics to the project and use them to instrument errors in reaching NATS and ServerService. Also observe and report latency metrics for the ConditionOrc API.

The prometheus scraper is managed by the same gin web server that hosts the API, so no need to add additional configuration.

### DIFF
--- a/pkg/api/v1/routes/handlers.go
+++ b/pkg/api/v1/routes/handlers.go
@@ -27,110 +27,92 @@ func (r *Routes) conditionKindValid(kind ptypes.ConditionKind) bool {
 	return found != nil
 }
 
-func (r *Routes) serverConditionUpdate(c *gin.Context) {
+func (r *Routes) serverConditionUpdate(c *gin.Context) (int, *v1types.ServerResponse) {
 	// XXX: should "uuid" be "server-uuid" or something?
 	serverID, err := uuid.Parse(c.Param("uuid"))
 	if err != nil {
-		c.JSON(
-			http.StatusBadRequest,
-			&v1types.ServerResponse{Message: "invalid ConditionUpdate payload " + err.Error()},
-		)
 		// XXX: do we have any way to correlate an API request id through this stack?
 		r.logger.WithFields(logrus.Fields{
 			"serverID": c.Param("uuid"),
 		}).Info("bad serverID")
-		return
+
+		return http.StatusBadRequest, &v1types.ServerResponse{
+			Message: "invalid ConditionUpdate payload " + err.Error(),
+		}
 	}
 
 	kind := ptypes.ConditionKind(c.Param("conditionKind"))
 	if !r.conditionKindValid(kind) {
-		c.JSON(
-			http.StatusBadRequest,
-			&v1types.ServerResponse{Message: "unsupported condition kind: " + string(kind)},
-		)
 		r.logger.WithFields(logrus.Fields{
 			"kind": kind,
 		}).Info("unsupported condition kind")
 
-		return
+		return http.StatusBadRequest, &v1types.ServerResponse{
+			Message: "unsupported condition kind: " + string(kind),
+		}
 	}
 
 	var conditionUpdate v1types.ConditionUpdate
 	if err = c.ShouldBindJSON(&conditionUpdate); err != nil {
-		c.JSON(
-			http.StatusBadRequest,
-			&v1types.ServerResponse{Message: "invalid ConditionUpdate payload " + err.Error()},
-		)
 		r.logger.WithFields(logrus.Fields{
 			"error": err,
 		}).Info("invalid ConditionUpdate")
 
-		return
+		return http.StatusBadRequest, &v1types.ServerResponse{
+			Message: "invalid ConditionUpdate payload " + err.Error(),
+		}
 	}
 
 	if conditionUpdate.ResourceVersion == 0 {
-		c.JSON(
-			http.StatusBadRequest,
-			&v1types.ServerResponse{Message: "invalid ConditionUpdate payload: resourceVersion not set"},
-		)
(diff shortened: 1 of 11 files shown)


pkg/api/v1/routes/handlers.go
```diff
@@ -322,140 +299,124 @@ func (r *Routes) exclusiveNonFinalConditionExists(ctx context.Context, serverID
...
 			"kind": kind,
 		}).Info("unsupported condition kind")
 
-		return
+		return http.StatusBadRequest, &v1types.ServerResponse{
+			Message: "unsupported condition kind: " + string(kind),
+		}
 	}
 
 	// TODO (vc): the repository Delete route should take a condition id
```
joelrebel, line 325: We need to discuss/ratify where and how the condition ID is to be used, its a bit blurry for me at the moment.

pkg/api/v1/routes/handlers.go
```diff
@@ -251,40 +224,44 @@ func (r *Routes) serverConditionCreate(c *gin.Context) {
...
 			"error": err,
 		}).Info("condition create failed")
 
-		return
+		return http.StatusInternalServerError, &v1types.ServerResponse{
+			Message: err.Error(),
+		}
 	}
 
+	// XXX: this operation can't ignore errors
```
joelrebel, line 258: My thoughts here was that the Condition `Orchestrator` component could follow up on `Condition`s that have no updates in the last X time period...

pkg/api/v1/routes/handlers.go
```diff
@@ -251,40 +224,44 @@ func (r *Routes) serverConditionCreate(c *gin.Context) {
 	// check if any condition with exclusive set is in incomplete states
 	if errEx := r.exclusiveNonFinalConditionExists(c.Request.Context(), serverID); errEx != nil {
 		if errors.Is(errEx, ErrConditionExclusive) {
-			c.JSON(
-				http.StatusBadRequest,
-				&v1types.ServerResponse{Message: errEx.Error()},
-			)
+			r.logger.WithFields(logrus.Fields{
+				"error": err,
+				// XXX: would be nice to have the id of the exclusive condition here too
```
joelrebel, line 229: :thumbsup

joelrebel (changes requested):

pkg/api/v1/routes/routes.go
```diff
@@ -63,6 +66,22 @@ func WithConditionDefinitions(defs ptypes.ConditionDefinitions) Option {
 	}
 }
 
+// apiHandler is a function that performs real work for the ConditionOrc API
```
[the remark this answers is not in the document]
DoctorVin: I don't know that I love the `gin-metrics` package, because I think our use-case is simpler than what this package provides. I don't see a need for tracking byte counts of requests and replies, for example. I think being explicit in our usage of prometheus gives us more flexibility in extending and adapting our usage in the future.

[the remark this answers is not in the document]
DoctorVin: One thing I *do* like about that package though, is that he uses a middle-ware paradigm instead of a function-wrapping paradigm. That might be nice to examine as an example hook for Auth/Az purposes later, or perhaps to refactor this.

pkg/api/v1/routes/handlers.go
DoctorVin: In light of our conversation this morning, this code is fine as-is. I'll remove that comment.

pkg/api/v1/routes/handlers.go
```diff
@@ -322,140 +299,124 @@ func (r *Routes) exclusiveNonFinalConditionExists(ctx context.Context, serverID
...
-		c.JSON(
-			http.StatusBadRequest,
-			&v1types.ServerResponse{Message: err.Error()},
-		)
-
-		return
+		return http.StatusBadRequest, &v1types.ServerResponse{Message: err.Error()}
 	}
 
+	// XXX: should this be condition_state in the gin Context?
```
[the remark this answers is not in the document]
DoctorVin: I thought this was a header or a JSON parameter. On second read I see that we're defining the token as `:conditionState` (and `:conditionKind` as well for other uses) so I agree, keeping things camelCase is better.

pkg/api/v1/routes/handlers.go
DoctorVin: I filed [issue 12](https://github.com/metal-toolbox/conditionorc/issues/12) to address this. This could be a simple "watcher" go-routine, or something more elaborate.

pkg/api/v1/routes/handlers.go
DoctorVin: cf. [issue 13](https://github.com/metal-toolbox/conditionorc/issues/13). I will likely take that on as a small follow up to this PR.

DoctorVin:
@joelrebel Let's discuss the pros/cons of adopting that third party metrics package? I don't have very strong feelings, but I would as soon not do it for testability reasons (that is, the current typed returns are easier to test rigorously) as well as simplicity reasons. I'm open to refactoring if you insist, though.

pkg/api/v1/routes/routes.go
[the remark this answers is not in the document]
joelrebel: yep, we do have a middleware that can we can use for Auth/Az later https://github.com/metal-toolbox/hollow-toolbox/tree/main/ginauth

joelrebel:
+1 for minimal external dependencies especially if we don't require most of the features from the external lib. 
I prefer the typed returns as well, lets move ahead with this and see how it goes.

joelrebel (approved):

PR #26, opened 2023-05-09
Pull controller updates from stream

joelrebel:
#### What does this PR do

- Implements changes to the orchestrator - to pull controller events from the Jetstream.
- Ack events and log ignored event data.
- Include an the condition ID field in conditionUpdate and validate before merge and commit to repository.

### DIFF
--- a/internal/orchestrator/orchestrator.go
+++ b/internal/orchestrator/orchestrator.go
@@ -206,16 +206,16 @@ func (o *Orchestrator) ackEvent(event events.Message, err error) {
 		}
 
 		// TODO: add metrics for dropped events
-		o.logger.WithError(err).WithFields(logFields).Error("event with error ack'ed")
+		o.logger.WithError(err).WithFields(logFields).Warn("event with error ack'ed")
 	}
 
 	if err := event.Ack(); err != nil {
-		o.logger.WithError(err).Error("error Ack'ing event")
+		o.logger.WithError(err).Warn("error Ack'ing event")
 	}
 }
 
 func (o *Orchestrator) eventNak(event events.Message) {
 	if err := event.Nak(); err != nil {
-		o.logger.WithError(err).Error("error Nack'ing event")
+		o.logger.WithError(err).Warn("error Nack'ing event")
 	}
 }


internal/orchestrator/orchestrator.go
```diff
@@ -89,56 +107,115 @@ func (o *Orchestrator) Run(ctx context.Context) {
...
+			logFields["event"] = data
+		} else {
+			logFields["event"] = event
 		}
 
-		o.logger.WithFields(
-			logrus.Fields{"err": err.Error(), "urn ns": urn.Namespace},
-		).Error("msg with unknown URN namespace ignored")
+		// TODO: add metrics for dropped events
+		o.logger.WithError(err).WithFields(logFields).Error("event with error ack'ed")
```
DoctorVin, line 209: do we need the `err: err.Error()` in `Fields` if we call the logger's `WithError` function? That seems redundant.

pkg/api/v1/events/handlers.go
```diff
@@ -167,6 +167,9 @@ func (h *Handler) updateCondition(ctx context.Context, streamEvent *ptypes.Strea
 		}).Info("condition merge failed")
 
 		// XXX: should this be an nak, instead of a ack.
+		//  - (joel) I don't think so, since this could be an out of order message and
+		// .  MergeExisting returns an error because the state transition cannot be done.
+		//    This could be a NakTerm instead so we can have metrics on it?
```
DoctorVin, line 172: I agree a `Term` would be appropriate here. On a Nak (in a perfect world 😅 ) the message will be redelivered, but the redelivery couldn't succeed, and if it did, it would be an error.

pkg/api/v1/types/types.go
```diff
@@ -112,6 +113,11 @@ func (c *ConditionUpdate) MergeExisting(existing *ptypes.Condition, compareResou
 		return nil, errBadUpdateTarget
 	}
 
+	if existing.ID != c.ID {
```
DoctorVin, line 116: I like this. It compels the controller to track the ID of the condition it is processing, but I don't think that's onerous.

internal/orchestrator/orchestrator.go
```diff
@@ -89,56 +107,115 @@ func (o *Orchestrator) Run(ctx context.Context) {
...
+
+	for _, msg := range msgs {
+		if ctx.Err() != nil || o.concurrencyLimit() {
+			o.eventNak(msg)
+
+			return
+		}
+
+		// spawn msg process handler
+		o.syncWG.Add(1)
```
DoctorVin, line 147: Nit/Food For Thought: shouldn't we increment by `len(msgs)`? Otherwise when the call to `PullMsg` is updated to `n>1` we'd have `concurrencyLimit` goroutines handling `len(msgs)` messages?

internal/orchestrator/orchestrator.go
```diff
@@ -89,56 +107,115 @@ func (o *Orchestrator) Run(ctx context.Context) {
...
+
+	data, err := event.Data()
+	if err != nil {
+		o.ackEvent(event, errors.Wrap(ErrInvalidEvent, "data field error: "+err.Error()))
 
 		return
 	}
 
-	urn, err := msg.SubjectURN(data)
+	urn, err := event.SubjectURN(data)
```
DoctorVin, line 171: Nit: What are examples of the components in the `SubjectURN`? I assume the UUID is the identifier for the server? But where do `namespace` and `prefix` come into play?

DoctorVin (dismissed):
Minor comments, nothing blocking a merge.

internal/orchestrator/orchestrator.go
joelrebel: `namespace` ends up in the [urn](https://github.com/infratographer/x/blob/main/urnx/types.go#L12),  its set to `hollow` in our [configuration](https://github.com/metal-toolbox/architecture/blob/f5b27686e634fed5b0215379f58418728e42227a/messaging.md#event-message-format), as of now its just conforming to what the other teams are following in their NATS messages. `prefix` is used to prefix the Subject when [publishing](https://github.com/metal-toolbox/hollow-toolbox/blob/0bfb6095e5a5566c0874968e1843fd8dddddfef8/events/nats.go#LL219C50-L219C72) the NATS event. Does that answer the query?

internal/orchestrator/orchestrator.go
joelrebel: I was trying to have the waitgroup `Add(1)` as close to the spawned routine just so returned errors don't cause an incorrect waitgroup increment. 
Can you elaborate on the second part of your question - `when the call to PullMsg is updated to n>1 we'd have concurrencyLimit goroutines handling len(msgs) messages?`

pkg/api/v1/types/types.go
joelrebel: 👍

internal/orchestrator/orchestrator.go
DoctorVin: As currently coded, the `PullMsg` method can pull a batch of messages created then 1. My question is: is this concurrency limit for limiting go-routines or messages-in-flight? What is the rationale for the limit?

internal/orchestrator/orchestrator.go
DoctorVin: It does, but it elicits more questions. I guess I don't understand what this URN is for. It doesn't conform to NATs subject rules (and even if it did, I'd object to its use because trying to debug a topic ID labeled with a UUID will be difficult). It seems like the SubjectURN refers to the target of the Condition message, so I'm trying to understand why we use URN instead of any other arbitrary identifier.

internal/orchestrator/orchestrator.go
joelrebel: Yes, the limit was introduced mainly so we can manage how many concurrent Goroutines are spawned, we can experiment with increasing it linearly based on how we see the Orchestrator handling messages.

DoctorVin (approved):
LGTM

PR #35, opened 2023-05-24
remove deprecated pubsubx package

DoctorVin:
This PR updates the code to remove the deprecated infratographer/pubsubx package. This was fairly invasive throughout the messaging layer for events. It also adds a `ConditionID` concept to conditions so that it is possible to track individual executions of an action. This also had some further-reaching changes because it affects the notion of what a `valid` condition or condition-update is.


pkg/api/v1/types/types.go
```diff
@@ -50,13 +50,20 @@ func (c *ConditionCreate) NewCondition(kind ptypes.ConditionKind) *ptypes.Condit
 
 // ConditionUpdate is the request payload to update an existing condition.
 type ConditionUpdate struct {
-	ID              uuid.UUID             `json:"id"`
+	ConditionID     uuid.UUID             `json:"condition_id"`
+	TargetID        uuid.UUID             `json:"target_id"`
```
joelrebel, line 54: TargetID here would be the object being updated, so the Server UUID in our case?

Could we stick to camelcase for the struct JSON tag values here

pkg/api/v1/events/handlers.go
```diff
@@ -39,228 +49,199 @@ func NewHandler(repository store.Repository, stream events.Stream, logger *logru
...
 
+	subject := msg.Subject()
+	fragment := strings.TrimPrefix(subject, controllerResponsePrefix)
+	//nolint:gomnd // useless, generic optinion
+	frags := strings.SplitN(fragment, ".", 2)
+
+	// XXX: this would be nicer if we could just destructure the output of SplitN,
+	// but golang is why we can't have nice things
+	facility := frags[0]
+	act := frags[1]
```
joelrebel, line 77: Worth having a len check on `frags` to make sure its got atleast two parts?

pkg/api/v1/events/handlers.go
```diff
@@ -39,228 +49,199 @@ func NewHandler(repository store.Repository, stream events.Stream, logger *logru
...
-		h.updateCondition(ctx, streamEvent)
+		var updateEvt v1types.ConditionUpdateEvent
+		if err := json.Unmarshal(msg.Data(), &updateEvt); err != nil {
+			h.logger.WithError(err).Warn("bogus condition update message")
+			h.ackEvent(msg)
+			return
+		}
+		err := h.updateCondition(ctx, &updateEvt)
+		switch {
+		case errors.Is(err, errRetryThis):
```
joelrebel, line 97: :thumbsup

pkg/api/v1/routes/handlers.go
```diff
@@ -136,6 +135,10 @@ func (r *Routes) serverConditionUpdate(c *gin.Context) (int, *v1types.ServerResp
 	return http.StatusOK, &v1types.ServerResponse{Message: "condition updated"}
 }
 
+// XXX: This needs to be refactored. In order to specifically compute the NATS subject for
```
joelrebel, line 138: The ServerID and conditionKind is known and should suffice to compute the NATS subject without necessarily caring about the json.RawMessage payload, no?

internal/orchestrator/orchestrator.go
```diff
@@ -158,53 +163,44 @@ func (o *Orchestrator) pullEvents(ctx context.Context) {
...
-	case ptypes.ControllerUrnNamespace:
-		o.eventHandler.ControllerEvent(otelCtx, streamEvent)
+	switch findSubjectOrigin(event.Subject()) {
+	case serverServiceOrigin:
+		o.eventHandler.ServerserviceEvent(otelCtx, event)
+	case controllersOrigin:
+		o.eventHandler.ControllerEvent(otelCtx, event)
 	default:
-		o.ackEvent(event, errors.Wrap(ErrInvalidEvent, "msg with unknown URN namespace ignored: "+urn.Namespace))
+		// how did we get a message delivered on a subject that we're not subscribed to?
```
joelrebel, line 194: stranger things...

joelrebel (changes requested):

pkg/types/types.go
```diff
@@ -188,6 +186,15 @@ type Fault struct {
 	FailAt string `json:"failAt,omitempty"`
 }
 
+// MustBytes returns an encoded json representation of the condition or panics
+func (c *Condition) MustBytes() []byte {
+	byt, err := json.Marshal(c)
+	if err != nil {
+		panic("encoding condition failed: " + err.Error())
```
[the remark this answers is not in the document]
DoctorVin: I'm not sure what you mean? This provides some syntactic sugar to avoid the clutter of dealing with a "should never happen" error every time we serialize a condition.

pkg/api/v1/events/handlers.go
DoctorVin: yeah, that's a defect on my part.

pkg/api/v1/routes/handlers.go
DoctorVin: This isn't possible in the current state, so I'm leaving it as-is and we'll address in follow-on work.

joelrebel (approved):
Thanks for the efforts here!

PR #45, opened 2023-06-01
Publish with facility code

joelrebel:
#### What does this PR do
- Adds a lookup for the facility code from the store
- Publishes created conditions on the NATS stream with the facility code in the subject.


DoctorVin (approved):
I'm ambivalent about adding a lookup in ConditionOrc for the facility when the client could (IMO should) send it in the request, but this works fine and we can address that after MVP is delivered.

PR #68, opened 2023-06-30
Tracing

joelrebel:
#### What does this PR do

- Instruments a few other methods for tracing
- Extracts trace identifiers from nats key values and links them for a complete overview

### DIFF
--- a/internal/orchestrator/updates.go
+++ b/internal/orchestrator/updates.go
@@ -12,6 +12,8 @@ import (
 	"github.com/metal-toolbox/conditionorc/internal/status"
 	v1types "github.com/metal-toolbox/conditionorc/pkg/api/v1/types"
 	ptypes "github.com/metal-toolbox/conditionorc/pkg/types"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/trace"
 
 	"github.com/nats-io/nats.go"
 	"github.com/pkg/errors"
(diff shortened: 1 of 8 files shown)


internal/orchestrator/updates.go
```diff
@@ -124,16 +130,33 @@ func parseStatusKVKey(key string) (*statusKey, error) {
...
 	MsgVersion int32           `json:"msgVersion"`
 }
 
+const (
+	Version int32 = 1
+)
+
+// MustBytes sets the version field of the StatusValue so any callers don't have
+// to deal with it. It will panic if we cannot serialize to JSON for some reason.
+func (v *flasherStatus) MustBytes() []byte {
```
DoctorVin, line 147: Do we need to publish a Flasher status from ConditionOrc?

DoctorVin (commented):
LGTM, just a question

internal/orchestrator/updates.go
joelrebel: not really, this was only added to cut the dependency on flasher in updates_test.go

joelrebel: I'm looking forward to moving all of these shared types into one place, then we have less duplication

DoctorVin: For the test, I'd rather actually explicitly rely on flasher (to avoid problems with flasher drifting and conditionorc being silently broken). It's only the test, and it's pretty manageable until we break that common eventing stuff out.

DoctorVin (approved):

PR #69, opened 2023-07-03
Define a few additional metrics

joelrebel:
#### What does this PR do

- Adds metrics and collection for conditions queued, completed and publish errors
- Moves metrics port to 9090

### DIFF
--- a/pkg/api/v1/routes/handlers.go
+++ b/pkg/api/v1/routes/handlers.go
@@ -10,6 +10,7 @@ import (
 	"github.com/gin-gonic/gin"
 	"github.com/google/uuid"
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


pkg/api/v1/routes/handlers.go
```diff
@@ -266,8 +278,23 @@ func (r *Routes) serverConditionCreate(c *gin.Context) (int, *v1types.ServerResp
 	}
 
 	// XXX: this operation can't ignore errors
+	// - if this method fails, should the condition be deleted?
```
DoctorVin, line 281: I think failing the condition is better here. That way we don't leave a caller confused as to why the condition they might have just seen (depending on how the UI/API interacts with serverservice) is now gone.

DoctorVin (approved):

pkg/api/v1/routes/handlers.go
joelrebel: Sounds right, will address this in another PR.

PR #72, opened 2023-07-10
reregister condition-orc with the liveness tracker

DoctorVin:
If we get an error on a liveness check-in, try to deregister/register this worker again. A failure on this re-registration is fatal.


joelrebel (changes requested):
Thanks for fixing this, can you include a counter metric for check in failures so that this can be tracked.

joelrebel (approved):

PR #73, opened 2023-08-03
Vc/kv updates only

DoctorVin:
#### What does this PR do
This PR removes the capability of ConditionOrc to listen for messages from controllers on a response channel in favor of only using the NATS KV store as a status tracking mechanism. This simplifies the code for extension; specifically it enables notification functionality in a cleaner way.


joelrebel (approved):
Thats a nice clean up, thanks!

PR #76, opened 2023-08-10
CreateCondition to revert new condition on failure

joelrebel:
#### What does this PR do

- When a newly created condition fails to get published on the event stream, the new condition is then purged,
and an error is returned to the client.

- Conditionorc client to format errors with returned status code.

### DIFF
--- a/pkg/api/v1/routes/handlers.go
+++ b/pkg/api/v1/routes/handlers.go
@@ -277,10 +277,7 @@ func (r *Routes) serverConditionCreate(c *gin.Context) (int, *v1types.ServerResp
 		}
 	}
 
-	// XXX: this operation can't ignore errors
-	// - if this method fails, should the condition be deleted?
-	//
-	// publish the condition
+	// publish the condition and in case of publish failure - revert.
 	err = r.publishCondition(otelCtx, serverID, facilityCode, condition)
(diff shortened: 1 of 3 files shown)


pkg/api/v1/routes/handlers.go
```diff
@@ -290,6 +287,21 @@ func (r *Routes) serverConditionCreate(c *gin.Context) (int, *v1types.ServerResp
 		metrics.PublishErrors.With(
 			prometheus.Labels{"conditionKind": string(condition.Kind)},
 		).Inc()
+
+		deleteErr := r.repository.Delete(otelCtx, serverID, condition.Kind)
+		if deleteErr != nil {
+			r.logger.WithFields(logrus.Fields{
+				"error": deleteErr,
+			}).Info("condition deletion failed")
```
DoctorVin, line 295: Nit: this should be a `Warn` so that it doesn’t get lost if someone filters logs.

DoctorVin (approved):
LGTM

pkg/api/v1/routes/handlers.go
joelrebel: For some reason all of the error logging is `Info` level, I'd like to look into that separately. Will leave this as is for now.

PR #81, opened 2023-08-16
restrict update KV subscription by the deployment facility

DoctorVin:
#### What does this PR do
We're in the process of moving deployment of the orchestrator functionality of ConditionOrc to the edge k8s clusters. To facilitate this notion of "orchestrator/flasher/alloy" as a unit of Fleet firmware services in a facility, we add configuration to restrict the keyspace of the condition data being subscribed to. This will mean that although Flasher is writing data to a "generic" KV bucket in NATS (called `firmwareInstall`), it does so with a key formatted as `facility.UUID` where the UUID is the ConditionID assigned to the `firmwareInstall` task.

The change proposed here restricts Orchestrators from viewing any update data whose key does not match a key the configured facility prefix.

####Note:
ConditionOrc should have a release cut after merge. This PR is marked _Draft_ to keep from merging prematurely. The code is ready for review.

### DIFF
--- a/internal/status/kv.go
+++ b/internal/status/kv.go
@@ -2,6 +2,7 @@ package status
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	ptypes "github.com/metal-toolbox/conditionorc/pkg/types"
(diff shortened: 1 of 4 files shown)


internal/status/kv.go
```diff
@@ -48,8 +49,8 @@ func ConnectToKVStores(s events.Stream, log *logrus.Logger, opts ...kv.Option) {
 
 // WatchFirmwareInstallStatus specializes some generic NATS functionality, mainly to keep
 // the callers cleaner of the NATS-specific details.
-func WatchFirmwareInstallStatus(ctx context.Context) (nats.KeyWatcher, error) {
-	// we can restrict the keys we watch (e.g. by facility code) here by using
-	// the KV Watch function instead.
-	return firmwareInstallKV.WatchAll(nats.Context(ctx))
+func WatchFirmwareInstallStatus(ctx context.Context, facility string) (nats.KeyWatcher, error) {
```
joelrebel, line 52: Could we stop this watcher once the Condition has been finalised? since we're going to be creating a lot more of these.

joelrebel (approved):

internal/status/kv.go
DoctorVin: The Watcher that is returned here watches the entire bucket, so this object handles all conditions coming in from the configured facility. We still only need one per instance.

joelrebel: Ah okay, I read it correctly now - for some reason I read `keyStr` to be `facility.ConditionID`

PR #88, opened 2023-08-31
Vc/dynamic kvs

DoctorVin:
#### What does this PR do

Instead of hard-coding a watcher for `FirmwareInstall` use the configured ConditionDefinitions to create the required watchers and manage them in a dynamic way. This paves the way for adding status monitoring for Alloy as well.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -71,7 +71,6 @@ linters:
     - misspell
     - noctx
     - stylecheck
-    - whitespace
     - gosec
   enable-all: false
   disable-all: true
--- a/internal/orchestrator/updates.go
+++ b/internal/orchestrator/updates.go
@@ -47,58 +47,109 @@ func (o *Orchestrator) startUpdateListener(ctx context.Context) {
 		if o.replicaCount > 1 {
 			opts = append(opts, kv.WithReplicas(o.replicaCount))
 		}
-		status.ConnectToKVStores(o.streamBroker, o.logger, opts...)
-		go o.statusKVListener(ctx)
+		status.ConnectToKVStores(o.streamBroker, o.logger, o.conditionDefs, opts...)
+		go o.kvStatusPublisher(ctx)
 	})
 }
 
-func (o *Orchestrator) statusKVListener(ctx context.Context) {
-	// start the watchers and return the associated channels
-	installWatcher, err := status.WatchFirmwareInstallStatus(ctx, o.facility)
-	if err != nil {
-		o.logger.WithError(err).Fatal("unable to watch install status KV")
-	}
-	// XXX: Alloy OOB support inventoryWatcher := status.WatchInventoryStatus(ctx)
-	o.logger.Info("listening for KV updates")
-	for {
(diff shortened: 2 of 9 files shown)


internal/orchestrator/updates.go
```diff
@@ -47,58 +47,109 @@ func (o *Orchestrator) startUpdateListener(ctx context.Context) {
...
-			le.WithError(err).Warn("updating condition")
+	o.logger.Debug("shut down KV updates")
+}
+
+type translatorFn func(context.Context, nats.KeyValueEntry) (*v1types.ConditionUpdateEvent, error)
+
+// startConditionListeners does what it says on the tin; iterate across all configured conditions
+// and start a KV watcher for each one. We increment the waitgroup counter for each condition we
+// can handle.
+func (o *Orchestrator) startConditionListeners(ctx context.Context,
```
joelrebel, line 102: Could we call this method `startConditionWatcher` instead?

internal/orchestrator/updates_test.go
```diff
@@ -146,3 +149,62 @@ func TestInstallEventFromKV(t *testing.T) {
...
+	ctx, cancel := context.WithCancel(context.TODO())
+
+	o.startConditionListeners(ctx, testChan, &wg)
+
+	sentinelChan := make(chan struct{})
+	toCtx, toCancel := context.WithTimeout(context.TODO(), time.Second)
+	defer toCancel()
+
+	var testPassed bool
+
```
joelrebel, line 194: Would be worth including a goroutine leak check in this test https://github.com/uber-go/goleak#quick-start

.golangci.yml
```diff
@@ -71,7 +71,6 @@ linters:
     - misspell
     - noctx
     - stylecheck
-    - whitespace
```
joelrebel, line 74: :'D

joelrebel (commented):
Thanks for the work here, looks good overall, left a few inline comments.

internal/orchestrator/updates.go
```diff
@@ -47,58 +47,109 @@ func (o *Orchestrator) startUpdateListener(ctx context.Context) {
...
+		wg.Add(1)
+
+		watcher, err := status.WatchConditionStatus(ctx, kind, o.facility)
+		if err != nil {
+			o.logger.WithError(err).WithField("condition.kind", string(kind)).Fatal("unable to get watcher")
 		}
+
+		go func() {
+			defer wg.Done()
+			// remember, the condition has to be true to run the loop, so "not stop" == false
```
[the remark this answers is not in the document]
DoctorVin: we can't use a break because it's in a `select` and that will take precedence (and do nothing). We could use a label and `goto` and I think that's as-ugly (plus potentially adds a headache for maintenance and extension).

internal/orchestrator/updates_test.go
DoctorVin: TIL. I'll look into it, because that's definitely something I want to avoid here.

joelrebel (approved):
🚀

PR #89, opened 2023-09-05
transform a key-value entry to a ConditionUpdate in a data-agnostic way

DoctorVin:
#### What does this PR do
This modifies the function that produces ConditionUpdateEvents from controller status. A key insight was that the `Kind` value was constant in the original function, but now can be parameterized.

### DIFF
--- a/internal/orchestrator/updates.go
+++ b/internal/orchestrator/updates.go
@@ -105,14 +105,6 @@ func (o *Orchestrator) startConditionWatchers(ctx context.Context,
 
 		kind := def.Kind
 
-		switch kind {
-		case ptypes.FirmwareInstall:
-		case ptypes.InventoryOutofband:
-		default:
-			o.logger.WithField("condition.kind", string(kind)).Warn("unsupported condition")
-			continue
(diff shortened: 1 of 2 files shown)


internal/orchestrator/updates.go
```diff
@@ -94,25 +94,20 @@ func (o *Orchestrator) kvStatusPublisher(ctx context.Context) {
...
 func (o *Orchestrator) startConditionWatchers(ctx context.Context,
 	evtChan chan<- *v1types.ConditionUpdateEvent, wg *sync.WaitGroup) {
 
 	for _, def := range o.conditionDefs {
-		var tf translatorFn
 		var watcher nats.KeyWatcher
 
 		kind := def.Kind
 
 		switch kind {
```
joelrebel, line 108: Can we drop this check as well, so its not limiting if we add new conditions or rename an existing one.

DoctorVin: As discussed, I'll remove this check.

joelrebel (approved):

PR #96, opened 2023-09-20
Vc/multistage build

DoctorVin:
#### What does this PR do
Adds a (currently experimental) multistage build so that we can more carefully control production builds (and still get the same caching benefits as we currently do). It also provides a simple way to override the docker image name. Finally, we add curl to the orchestrator image to assist in debugging some Slack notification issues we're seeing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ VERSION     := $(shell git describe --tags 2> /dev/null)
 BUILD_DATE  := $(shell date +%s)
 GIT_COMMIT_FULL  := $(shell git rev-parse HEAD)
 GO_VERSION := $(shell expr `go version |cut -d ' ' -f3 |cut -d. -f2` \>= 16)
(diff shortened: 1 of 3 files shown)


Alva8756 (approved):

Makefile
```diff
@@ -7,7 +7,7 @@ VERSION     := $(shell git describe --tags 2> /dev/null)
 BUILD_DATE  := $(shell date +%s)
 GIT_COMMIT_FULL  := $(shell git rev-parse HEAD)
 GO_VERSION := $(shell expr `go version |cut -d ' ' -f3 |cut -d. -f2` \>= 16)
-DOCKER_IMAGE  := "ghcr.io/metal-toolbox/conditionorc"
+DOCKER_IMAGE  ?= "ghcr.io/metal-toolbox/conditionorc"
```
jakeschuurmans, line 10: Question of curiousity: Can DOCKER_IMAGE sometimes be set outside of the makefile? Or is this just to allow that for testing purposes?

PR #116, opened 2023-10-05
Adding conditions and server id to gin endpoint tracing

jakeschuurmans:
#### What does this PR do

Add additional information (server id and condition state/kind) to jaeger traces


jakeschuurmans:
Only question I have about this is if we are happy with my naming convention. "conditionKind" for example, should it be camelCase? Or snake_case? Or flatcase?

joelrebel:
yeah... for now lets leave it as is, since thats the convention followed in all other components. We can take up switching to `snake_case` separately - when we see its a pain to query

DoctorVin (approved):

PR #117, opened 2023-10-09
take multi-token auth config instead of individual variables

DoctorVin:
#### What does this PR do
This PR changes condition-api from being configured to accept OIDC authentication via environment variables to being configured via the Kubernetes config-map (mounted as a configuration file).

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -91,5 +91,6 @@ var cmdServer = &cobra.Command{
 // install command flags
 func init() {
 	rootCmd.AddCommand(cmdServer)
-	ginjwt.RegisterViperOIDCFlags(viper.GetViper(), cmdServer)
+	cmdServer.Flags().Bool("oidc", true, "use oidc auth")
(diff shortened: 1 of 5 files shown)


cmd/server.go
```diff
@@ -56,7 +58,13 @@ var cmdServer = &cobra.Command{
 			server.WithStore(repository),
 			server.WithStreamBroker(streamBroker),
 			server.WithConditionDefinitions(app.Config.ConditionDefinitions),
-			server.WithAuthMiddlewareConfig(app.Config.APIServerJWTAuth),
+		}
+
+		if viper.GetViper().GetBool("oidc.enabled") {
```
joelrebel, line 63: Can we keep the configuration loading within the app package https://github.com/metal-toolbox/conditionorc/blob/ce8c4a67b0be44b411a40c959666e2ea8f58d395/internal/app/config.go#L118

DoctorVin:
I tested this by overriding a pod in `CH3-Hollow` and directing some `mctl` traffic at it. It worked as expected, so I believe this change is safe and will accomplish the desired effect.

joelrebel (approved):
thanks

PR #121, opened 2023-10-19
Vc/nats condition tracking

DoctorVin:
#### What does this PR do
Add a NATS kv-based store for conditions. This simplifies tracking active conditions and reduces condition-api's and condition-orchestrator's dependence on server-service.

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -32,17 +32,17 @@ var cmdServer = &cobra.Command{
 			log.Fatal(err)
 		}
 
-		repository, err := store.NewStore(cmd.Context(), app.Config, app.Config.ConditionDefinitions, app.Logger)
-		if err != nil {
-			app.Logger.Fatal(err)
-		}
-
 		streamBroker, err := events.NewStream(app.Config.NatsOptions)
 		if err != nil {
 			app.Logger.Fatal(err)
 		}
 
 		if err := streamBroker.Open(); err != nil {
+			app.Logger.Fatal(err)
+		}
+
+		repository, err := store.NewStore(app.Config, app.Config.ConditionDefinitions, app.Logger, streamBroker)
+		if err != nil {
 			app.Logger.Fatal(err)
 		}
 
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -37,11 +37,14 @@ var cmdServer = &cobra.Command{
 			app.Logger.Fatal(err)
 		}
 
-		if err := streamBroker.Open(); err != nil {
+		if err = streamBroker.Open(); err != nil {
 			app.Logger.Fatal(err)
 		}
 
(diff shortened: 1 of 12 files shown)


internal/store/interface.go
```diff
@@ -51,11 +53,99 @@ type Repository interface {
...
 	default:
 		return nil, errors.Wrap(ErrRepository, "storage kind not implemented: "+string(config.StoreKind))
 	}
 }
+
+func getServerServiceClient(ctx context.Context, cfg *app.ServerserviceOptions, log *logrus.Logger) (*sservice.Client, error) {
+	var client *sservice.Client
+	var err error
+
+	if !cfg.DisableOAuth {
```
joelrebel, line 84: this conditional could be inverted for readability
```
  if cfg.DisableOAuth {
    return newClientWithOAuth(ctx, cfg, log)
  }

  return sservice.NewClientWithToken("fake", cfg.Endpoint, nil)
```

cmd/server.go
```diff
@@ -32,17 +32,20 @@ var cmdServer = &cobra.Command{
...
-		if err != nil {
+		if err = streamBroker.Open(); err != nil {
 			app.Logger.Fatal(err)
 		}
 
-		if err := streamBroker.Open(); err != nil {
+		// setup cancel context with cancel func
+		ctx, serverCancel := context.WithCancel(cmd.Context())
+
+		repository, err := store.NewStore(ctx, app.Config, app.Config.ConditionDefinitions, app.Logger, streamBroker)
```
joelrebel, line 47: We require two separate repository instances available here - one for the Condition and the other for HSS server data?
That would mean `store.NewStore` has to be invoked once for each of them,
```
conditionStore, err := store.NewStore(ctx, app.Config, app.Config.ConditionDefinitions, app.Logger, streamBroker)
...
serverStore, err := store.NewStore(ctx, app.Config, app.Config.ConditionDefinitions, app.Logger, nil)
```

And we'd want to rename and create `Config.StoreKind` as two separate fields in the configuration,

https://github.com/metal-toolbox/conditionorc/blob/6d0c0e86b8508e4ba82b5301137bfe3f197e4ec2/internal/app/config.go#L39
```
StoreKindConditions model.StoreKind   
StoreKindServers model.StoreKind
```

joelrebel (changes requested):

cmd/server.go
DoctorVin: We don’t need 2 instances of `Store`. We either get one specialized to NATS or Serverservice. Both incorporate a client for HSS (for now), but I intentionally bury that implementation detail. The HSS client is only for server lookup.

joelrebel: Ignore my comment here, I was missing the fact that the NATs store wrapped the HSS client in there.

joelrebel (approved):

joelrebel (approved):

PR #128, opened 2023-10-26
remove client support for dead apis

DoctorVin:
#### What does this PR do

This catches the client up with the current state of the API handlers.


mmlb (approved):
testing. 1, 2, 3.

PR #130, opened 2023-10-27
Vc/integrate new store

DoctorVin:
#### What does this PR do

This represents some changes and updates from integrating the NATS-based condition store in the sandbox. Once approved, we can start testing things in production.


internal/store/nats.go
```diff
@@ -174,9 +175,58 @@ func (n *natsStore) GetServer(ctx context.Context, serverID uuid.UUID) (*model.S
...
+			// is enough to signal to the API handler that it should *not* accept a new condition
+			// for this server
+			outgoing.Exclusive = true
+			outgoing.State = incState
+			break
+		}
+	}
+
+	if !found {
+		return nil, ErrConditionNotFound
```
joelrebel, line 224: Why does this have to be an error - if no conditions are found on a server?

internal/store/nats.go
```diff
@@ -174,9 +175,58 @@ func (n *natsStore) GetServer(ctx context.Context, serverID uuid.UUID) (*model.S
 	return &model.Server{ID: obj.UUID, FacilityCode: obj.FacilityCode}, nil
 }
 
-// List is deprecated and is not implemented here
-func (n *natsStore) List(_ context.Context, _ uuid.UUID, _ rctypes.State) ([]*rctypes.Condition, error) {
-	return nil, errors.New("unimplemented")
+// List retrieves any active conditions
+func (n *natsStore) List(ctx context.Context, srvID uuid.UUID, incState rctypes.State) ([]*rctypes.Condition, error) {
```
joelrebel, line 179: should we call this method `ListActive()` so its role is clearer upfront?
And since the method only returns the first found active condition, I wonder if it should be called something else altogether.

internal/store/nats.go
```diff
@@ -174,9 +175,58 @@ func (n *natsStore) GetServer(ctx context.Context, serverID uuid.UUID) (*model.S
...
+		return nil, errors.Wrap(ErrRepository, err.Error())
+	}
+
+	var found bool
+	var outgoing rctypes.Condition
+	for _, c := range cr.Conditions {
+		//nolint:gocritic // inverting this test is a moronic suggestion and you should feel bad about it
+		if !rctypes.StateIsComplete(c.State) {
+			// The first incomplete condition is good enough
+			found = true
```
joelrebel, line 211: instead of assigning and breaking out here, could we `return  []*rctypes.Condition{&outgoing}, nil`

internal/store/nats.go
```diff
@@ -174,9 +175,58 @@ func (n *natsStore) GetServer(ctx context.Context, serverID uuid.UUID) (*model.S
...
+			outgoing = *c
+			// make sure to match the requested parameters to short circuit the loops in the handler.
+			// we don't care what the actual state of the found condition is; the fact that it's incomplete
+			// is enough to signal to the API handler that it should *not* accept a new condition
+			// for this server
+			outgoing.Exclusive = true
```
joelrebel, line 217: why does `Exclusive` have to be set here?

pkg/api/v1/routes/handlers.go
```diff
@@ -241,6 +241,8 @@ func (r *Routes) serverFacilityCode(ctx context.Context, serverID uuid.UUID) (st
 func (r *Routes) exclusiveNonFinalConditionExists(ctx context.Context, serverID uuid.UUID) error {
 	otelCtx, span := otel.Tracer(pkgName).Start(ctx, "Routes.exclusiveNonFinalConditionExists")
 	defer span.End()
+	// XXX: Clean me up! We're looking for any active condition here, so don't list, do a store lookup
+	// here. Any active condition here is enough to block queuing a new one.
```
joelrebel, line 245: Note: we'd only want active `Exclusive` conditions to block queueing a new one.

joelrebel (commented):
Left a few comments on there.

internal/store/nats.go
DoctorVin: I’m matching [existing behavior](https://github.com/metal-toolbox/conditionorc/blob/42b392243ba12cd425c3583e59d92978e0554f35/internal/store/serverservice.go#L64)

internal/store/nats.go
DoctorVin: Right now, this function needs to implement the existing Repository [interface](https://github.com/metal-toolbox/conditionorc/blob/42b392243ba12cd425c3583e59d92978e0554f35/internal/store/interface.go#L24)

internal/store/nats.go
DoctorVin: We can’t queue work to be done concurrently without exhaustively analysis that they won’t conflict.

internal/store/nats.go
DoctorVin: The current handler inspects the condition, because of the way Serverservice handles conditions (that is, as versioned attributes). The handler is too smart here, but we can’t address that until we can swap NATS in for the other repository.

pkg/api/v1/routes/handlers.go
DoctorVin: All actions should be exclusive, especially once we start queuing a group of them at once.

internal/store/nats.go
joelrebel: Thats the `Get` method, the `List` will continue on a `404` https://github.com/metal-toolbox/conditionorc/blob/42b392243ba12cd425c3583e59d92978e0554f35/internal/store/serverservice.go#L131

joelrebel (approved):

PR #132, opened 2023-10-29
FS-996: server enrollment API endpoint

Alva8756:
* creates a server record in FleetDB
* schedules an `inventory` condition on the device


pkg/api/v1/routes/handlers.go
```diff
@@ -79,8 +79,54 @@ func (r *Routes) serverConditionCreate(c *gin.Context) (int, *v1types.ServerResp
...
+	// Creates a server record in FleetDB.
+	if err = r.repository.AddServer(c.Request.Context(), serverID, c.Param("ip"), c.Param("user"), c.Param("pwd")); err != nil {
+		return http.StatusBadRequest, &v1types.ServerResponse{
+			Message: err.Error(),
+		}
+	}
+
+	// Reuse the inventory for server enrollment.
+	// May change in the future.
+	kind := rctypes.Kind("inventory")
```
joelrebel, line 113: Use the constant from the rivets library https://github.com/metal-toolbox/rivets/blob/e4c2952b8cb4530334d7a4135f0dc994376375e6/condition/inventory.go#L11

pkg/api/v1/routes/handlers.go
```diff
@@ -79,8 +79,54 @@ func (r *Routes) serverConditionCreate(c *gin.Context) (int, *v1types.ServerResp
 		}
 	}
 
+	return r.conditionCreate(otelCtx, c, kind, serverID)
+}
+
+func (r *Routes) serverEnroll(c *gin.Context) (int, *v1types.ServerResponse) {
```
joelrebel, line 85: Can you include tests for this endpoint

pkg/api/v1/routes/handlers.go
```diff
@@ -79,8 +79,54 @@ func (r *Routes) serverConditionCreate(c *gin.Context) (int, *v1types.ServerResp
...
+		}).Info("bad serverID")
+
+		return http.StatusBadRequest, &v1types.ServerResponse{
+			Message: err.Error(),
+		}
+	}
+
+	// Creates a server record in FleetDB.
+	if err = r.repository.AddServer(c.Request.Context(), serverID, c.Param("ip"), c.Param("user"), c.Param("pwd")); err != nil {
+		return http.StatusBadRequest, &v1types.ServerResponse{
```
joelrebel, line 106: Check if the error returned by HSS/Fleetdb contains a duplicate key error or similar (which indicates the server already exists) then return `http.StatusBadRequest`,
for other errors return `http.StatusInternalServerError`

internal/store/serverservice.go
```diff
@@ -179,6 +180,32 @@ func (s *Serverservice) Create(ctx context.Context, serverID uuid.UUID, createCo
 	return err
 }
 
+// AddServer creates a server record in FleetDB
+// @id: required
+// @ip: required
+// @user: required
+// @pwd: required
+func (s *Serverservice) AddServer(ctx context.Context, serverID uuid.UUID, bmcAddr, bmcUser, bmcPass string) error {
```
joelrebel, line 188: This needs to accept the facility code parameter as well, each server must have a facility code.

pkg/api/v1/routes/handlers.go
```diff
@@ -79,8 +79,54 @@ func (r *Routes) serverConditionCreate(c *gin.Context) (int, *v1types.ServerResp
...
+
+		return http.StatusBadRequest, &v1types.ServerResponse{
+			Message: "unsupported condition kind: " + string(kind),
+		}
+	}
+
+	return r.conditionCreate(otelCtx, c, kind, serverID)
+}
+
+func (r *Routes) conditionCreate(otelCtx context.Context, c *gin.Context, kind rctypes.Kind, serverID uuid.UUID) (int, *v1types.ServerResponse) {
```
joelrebel, line 127: I suggest we split this `serverConditionCreate` function, leave the previous name as is and create another function `conditionCreate` with contents from line https://github.com/metal-toolbox/conditionorc/blob/ed108837070cc103684754430d0b8383a72a7ec8/pkg/api/v1/routes/handlers.go#L213

This is because we don't require to perform any validation that the current `serverConditionCreate` method does - since this is new server.

internal/store/interface.go
```diff
@@ -49,6 +49,13 @@ type Repository interface {
 	// @serverID: required
 	// @conditionKind: required
 	Delete(ctx context.Context, serverID uuid.UUID, conditionKind rctypes.Kind) error
+
+	// AddServer creates a server record in FleetDB
+	// @serverID: required
+	// @bmcAddr: required
+	// @bmcUser: required
+	// @bmcPass: required
+	AddServer(ctx context.Context, serverID uuid.UUID, bmcAddr string, bmcUser string, bmcPass string) error
```
DoctorVin, line 58: We might want a different interface here for managing the server population. Conditions are moving to a NATS backed system and I am (was?) intending on deleting the Server-service condition store implementation.

internal/store/nats.go
```diff
@@ -211,6 +211,11 @@ func (n *natsStore) Create(ctx context.Context, serverID uuid.UUID, condition *r
 	return err
 }
 
+// AddServer creates a server record in FleetDB
```
DoctorVin, line 214: In line with my comment above, we're never going to use NATS to store servers so we should avoid implementing even this much in the NATS repo.

internal/store/serverservice.go
```diff
@@ -179,6 +180,32 @@ func (s *Serverservice) Create(ctx context.Context, serverID uuid.UUID, createCo
...
+	// Add server
+	server := sservice.Server{UUID: serverID, Name: serverID.String()}
```
DoctorVin, line 190: does the ServerService API have a way to pack this all into a single transaction? Catching an error mid-way and aborting all these individual updates is going to be tedious and erorr-prone.

internal/store/serverservice.go
```diff
@@ -179,6 +180,32 @@ func (s *Serverservice) Create(ctx context.Context, serverID uuid.UUID, createCo
...
+
+	// Add server BMC credential
+	_, err = s.client.SetCredential(ctx, serverID, "bmc", bmcUser, bmcPass)
+	if err != nil {
+		return err
+	}
+
+	// Add server BMC IP attribute
+	addrAttr := fmt.Sprintf(`{"address": "%q"}`, bmcAddr)
+	bmcIPAttr := sservice.Attributes{Namespace: "sh.hollow.bmc_info", Data: []byte(addrAttr)}
```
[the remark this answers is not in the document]
DoctorVin: @joelrebel Do we have other tools currently (or planned) that would update the BMC info? If not it's probably better to keep this private.

pkg/api/v1/routes/handlers.go
```diff
@@ -79,8 +79,54 @@ func (r *Routes) serverConditionCreate(c *gin.Context) (int, *v1types.ServerResp
...
+	if err = r.repository.AddServer(c.Request.Context(), serverID, c.Param("ip"), c.Param("user"), c.Param("pwd")); err != nil {
+		return http.StatusBadRequest, &v1types.ServerResponse{
+			Message: err.Error(),
+		}
+	}
+
+	// Reuse the inventory for server enrollment.
+	// May change in the future.
+	kind := rctypes.Kind("inventory")
+	if !r.conditionKindValid(kind) {
```
DoctorVin, line 114: You don't need to test the validity of the kind here; we're not accepting untrusted external input. We're specifically using a valid condition `Kind`.

pkg/api/v1/routes/handlers.go
DoctorVin: Likewise, the `gin.Context` is not going to have a POST payload that fits here. I agree with Joel's comment that we should pull the common `createCondition` elements into a new, un-exported function and call that from `serverEnroll`.

Don't go too crazy updating the existing `conditionCreate`, I will be making a bunch of changes in that area.

internal/store/errors.go
```diff
@@ -3,10 +3,11 @@ package store
...
-	ErrMalformedCondition   = errors.New("condition data is invalid")
-	ErrNoCreate             = errors.New("unable to create condition")
-	ErrConditionNotComplete = errors.New("condition is not complete") // can't delete
+	ErrServerNotFound          = errors.New("server not found")
+	ErrConditionExists         = errors.New("condition exists on server")
+	ErrConditionNotFound       = errors.New("condition not found")
+	ErrMalformedCondition      = errors.New("condition data is invalid")
+	ErrNoCreate                = errors.New("unable to create condition")
+	ErrConditionNotComplete    = errors.New("condition is not complete") // can't delete
+	ErrAddServerNotImplemented = errors.New("add-server not implemented")
```
DoctorVin, line 12: We don't need to export this error, see below.

DoctorVin (changes requested):

pkg/api/v1/routes/handlers.go
joelrebel: > Do we have other tools currently (or planned) that would update the BMC info? If not it's probably better to keep this private.
@DoctorVin we do have systems that require to look up that data, and must know the namespace to look it up from.

pkg/api/v1/routes/handlers.go
```diff
@@ -79,8 +79,54 @@ func (r *Routes) serverConditionCreate(c *gin.Context) (int, *v1types.ServerResp
 		}
 	}
 
+	return r.conditionCreate(otelCtx, c, kind, serverID)
```
joelrebel, line 82: Note: the `inventory` condition needs to be created with certain parameters - you can follow the same pattern as in mctl - https://github.com/metal-toolbox/mctl/blob/fd42322cb93842dc993dee44d70f986230b9e6c0/cmd/collect/inventory.go#L46

pkg/api/v1/routes/handlers.go
Alva8756: Thanks for the suggetstion Joel and Vince!
I simply wrap codes starting from line 213 into a function following Joels suggestions in order to bypass the unnecessary validation checks.
PTAL.

internal/store/interface.go
Alva8756: I am not familiar with the future plan about refactoring of the `Conditionorc` repo thus may need some insights how we want to scope the new interface.
As a temporary workaround, I am making a public function which receives the `*ServerService` store as a parameter.

pkg/api/v1/routes/handlers.go
Alva8756: Thanks Jeol!
It sounds basically I am understanding the ticket correctly?
I will start adding tests and verifying it works in sandbox

pkg/api/v1/routes/handlers.go
Alva8756: Thanks! Do you mean I can reuse exactly the same pattern for new added `add server`? Since we are reusing the `inventory` kind

pkg/api/v1/routes/handlers.go
```diff
@@ -163,8 +163,71 @@ func (r *Routes) serverConditionCreate(c *gin.Context) (int, *v1types.ServerResp
...
+	defer span.End()
+
+	repo, ok := r.repository.(*store.Serverservice)
+	if !ok {
+		return http.StatusInternalServerError, &v1types.ServerResponse{
+			Message: "add-server only support server-service store",
+		}
+	}
+
+	serverID, err := uuid.Parse(c.Param("uuid"))
```
DoctorVin, line 184: You have two parameters referencing the same concept: `id` (line 172) and `uuid` (here). I suggest 2 things:

1. Make the parameter name `serverID` in the POST payload (that is, what comes through on the `*gin.Context`)
2. Handle this as an optional parameter. If it's missing, create one via `uuid.New()`

pkg/api/v1/routes/handlers.go
```diff
@@ -163,8 +163,71 @@ func (r *Routes) serverConditionCreate(c *gin.Context) (int, *v1types.ServerResp
...
+		}).Info("bad serverID")
+
+		return http.StatusBadRequest, &v1types.ServerResponse{
+			Message: err.Error(),
+		}
+	}
+
+	facilityCode := c.Param("facility")
+	// Creates a server record in FleetDB.
+	if err = store.AddServer(c.Request.Context(), repo, serverID, c.Param("facility"), c.Param("ip"), c.Param("user"), c.Param("pwd")); err != nil {
```
DoctorVin, line 197: We should pull out/sanity check our parameters up front in the function. We keep calling `c.Param(token)` and that's annoying.

pkg/api/v1/routes/handlers.go
```diff
@@ -163,8 +163,71 @@ func (r *Routes) serverConditionCreate(c *gin.Context) (int, *v1types.ServerResp
...
+			Message: err.Error(),
+		}
+	}
+
+	// Reuse the inventory for server enrollment.
+	// May change in the future.
+	kind := rctypes.Kind(rctypes.Inventory)
+
+	var conditionCreate v1types.ConditionCreate
+	if err := c.ShouldBindJSON(&conditionCreate); err != nil {
```
DoctorVin, line 213: Calling `ShouldBindJSON` here assumes that we're going to add all of the applicable contents for `conditionCreate` in the POST payload. Instead, we should define standard `inventory` parameters here (specify `outofband` at least, other things might be applicable) and use that to craft your new Condition.

DoctorVin (changes requested):

internal/store/interface.go
DoctorVin: In short, I had been planning on removing `internal/store/serverservice.go`. That said, we might want to provide a nice interface to FleetDB to get various database objects. We could create a `Servers` interface for FleetDB that encapsulated the client and provided an API that was less "raw" for manipulating objects in the Servers table (and associated attributes).

internal/store/serverservice.go
```diff
@@ -179,6 +186,44 @@ func (s *Serverservice) Create(ctx context.Context, serverID uuid.UUID, createCo
...
+	addr, err := netip.ParseAddr(bmcAddr)
+	if err != nil {
+		return err
+	}
+
+	// Add server
+	server := sservice.Server{UUID: serverID, Name: serverID.String(), FacilityCode: facilityCode}
+	_, _, err = s.client.Create(ctx, server)
+	if err != nil {
+		if strings.Contains(err.Error(), serverExistErrMsg) {
```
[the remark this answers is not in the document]
Alva8756: Oh I was following the other comment: https://github.com/metal-toolbox/conditionorc/pull/132#discussion_r1376050796

Basically `Check if the error returned by HSS/Fleetdb contains a duplicate key error or similar (which indicates the server already exists) then return http.StatusBadRequest, for other errors return http.StatusInternalServerError`.

I don't have preference on it since `mtcl` will see the same error message which users can tell if the id is duplicated or not. Maybe `BadRequest` is appropriate?

I don't have a good way besides reading the error message. I don't find useful information inside `ServerResponse`

internal/store/serverservice.go
DoctorVin: OK, let's do it the hard way to start then. But we will need a method for ourselves that removes a broken server record and gets things cleaned up.

DoctorVin (changes requested):
I like what I see here, and I think it's a good start to what we need to do. However I'd like to see the code that wires in the routes disabled for now. My preference is that your code get into `main` sooner rather than later, but I don't want it to be executed in production until we're ready for it.

For next steps, I think we need to work through a few things:

1. We need test coverage. There is a lot of logic here and we need to structure it so that we can unit-test appropriately before we start integration.
2. We need to do some work around parameter handling. right now we're mixing calls to `gin.Context.Param` and `gin.Context.ShouldBindJSON`. The calls to `Param()` are only applicable for parameters passed on the URL (which is mainly `GET` requests), but we're handling a `POST` to the `serverEnroll` endpoint. I have some specific ideas we can discuss when you get to this point. You can take a look at the gin docs for [Param](https://pkg.go.dev/github.com/gin-gonic/gin#Context.Param) and [ShouldBindJSON](https://pkg.go.dev/github.com/gin-gonic/gin#Context.ShouldBindJSON)

so `tl;dr` I want a few things, but this is the last for this PR.

Alva8756:
> However I'd like to see the code that wires in the routes disabled for now.

Hi @DoctorVin , by `wires in the routes disabled for now`, do you mean we submit the code but the new feature is disabled by default?
If so, we can introduce a flag or env variable for registering the new handler. It will be only enabled by tests until we are ready for rolloiut?

Alva8756:
> I like what I see here, and I think it's a good start to what we need to do. However I'd like to see the code that wires in the routes disabled for now. My preference is that your code get into `main` sooner rather than later, but I don't want it to be executed in production until we're ready for it.
> 
> For next steps, I think we need to work through a few things:
> 
> 1. We need test coverage. There is a lot of logic here and we need to structure it so that we can unit-test appropriately before we start integration.
> 2. We need to do some work around parameter handling. right now we're mixing calls to `gin.Context.Param` and `gin.Context.ShouldBindJSON`. The calls to `Param()` are only applicable for parameters passed on the URL (which is mainly `GET` requests), but we're handling a `POST` to the `serverEnroll` endpoint. I have some specific ideas we can discuss when you get to this point. You can take a look at the gin docs for [Param](https://pkg.go.dev/github.com/gin-gonic/gin#Context.Param) and [ShouldBindJSON](https://pkg.go.dev/github.com/gin-gonic/gin#Context.ShouldBindJSON)
> 
> so `tl;dr` I want a few things, but this is the last for this PR.

Hi @DoctorVin ,
I moved some params from URL to payload. Leaving only `uuid` in the URL. Does that sound good to you? Thanks!

pkg/api/v1/routes/handlers.go
```diff
@@ -163,8 +168,74 @@ func (r *Routes) serverConditionCreate(c *gin.Context) (int, *v1types.ServerResp
...
+			}
+		}
+		return http.StatusInternalServerError, &v1types.ServerResponse{
+			Message: err.Error(),
+		}
+	}
+
+	// Reuse the inventory for server enrollment.
+	// May change in the future.
+	kind := rctypes.Kind(rctypes.Inventory)
```
DoctorVin, line 229: You don't need to cast `rctypes.Inventory` as a `rctypes.Kind`.

pkg/api/v1/routes/handlers.go
```diff
@@ -163,8 +168,74 @@ func (r *Routes) serverConditionCreate(c *gin.Context) (int, *v1types.ServerResp
...
+
+	newCondition := conditionCreate.NewCondition(kind)
```
DoctorVin, line 231: At this point, the `Parameters` field of newCondition is going to be the payload we bound at line 181, right? So you're creating an `Inventory` condition with `AddServer` parameters. What we need to do here is create a [InventoryTaskParameters](https://github.com/metal-toolbox/rivets/blob/e4c2952b8cb4530334d7a4135f0dc994376375e6/condition/inventory.go#L21) and serialize that. This would be something like:
```
inventoryArgs := rctypes.{
    CollectBiosConfig: true,
    CollectFirmwareStatus: true,
    Method: rctypes.Outofband,
    AssetID: serverID,
}
``` 
then serialize that to JSON and set it as the `Parameters` in newCondition.

DoctorVin (changes requested):

internal/store/serverservice.go
Alva8756: Sounds good to me. I will create a ticket and add `DeleteServer` as a follow up PR.

There might be some details I need to think of, eg, if we fails deleting a server which is created successfully, we may want to retry deleting until it is deleted?

Anyway, it may worth a separate PR for cleaning up?

pkg/api/v1/routes/handlers.go
Alva8756: Thanks!
The current version filtered out params for creating the server. PTAL thanks!

PR #137, opened 2023-11-07
[FS-1019]Add delete cleanup feature to handle server enroll cleanup

Alva8756:
This PR add delete cleanup feature to handle server enroll cleanup
to address the comment: https://github.com/metal-toolbox/conditionorc/pull/132#discussion_r1379141832

### DIFF
--- a/internal/fleetdb/fleetdb.go
+++ b/internal/fleetdb/fleetdb.go
@@ -20,11 +20,21 @@ import (
 	sservice "go.hollow.sh/serverservice/pkg/api/v1"
 )
 
+const (
+	secretSlug = "bmc"
+)
+
 type fleetDBImpl struct {
 	config               *app.ServerserviceOptions
 	conditionDefinitions rctypes.Definitions
 	client               *sservice.Client
 	logger               *logrus.Logger
+}
+
+type serverCreateStatus struct {
+	serverCreated     bool
+	credentialCreated bool
+	attributesCreated bool
 }
 
 var (
@@ -40,37 +50,68 @@ func serverServiceError(operation string) {
 }
 
 // AddServer creates a server record in FleetDB
-func (s *fleetDBImpl) AddServer(ctx context.Context, serverID uuid.UUID, facilityCode, bmcAddr, bmcUser, bmcPass string) error {
+func (s *fleetDBImpl) AddServer(ctx context.Context, serverID uuid.UUID, facilityCode, bmcAddr, bmcUser, bmcPass string) (func() error, error) {
+	var createStatus serverCreateStatus
+	cleanup := func() error {
+		if createStatus.serverCreated {
+			server := sservice.Server{UUID: serverID, Name: serverID.String(), FacilityCode: facilityCode}
+			_, err := s.client.Delete(ctx, server)
+			if err != nil {
+				return err
+			}
+		}
+
+		if createStatus.credentialCreated {
+			_, err := s.client.DeleteCredential(ctx, serverID, secretSlug)
+			if err != nil {
+				return err
+			}
+		}
+
+		if createStatus.attributesCreated {
+			_, err := s.client.DeleteAttributes(ctx, serverID, rservice.ServerAttributeNSBmcAddress)
+			if err != nil {
+				return err
+			}
+		}
+		return nil
+	}
 	otelCtx, span := otel.Tracer(pkgName).Start(ctx, "FleetDB.AddServer")
 	defer span.End()
 
 	if bmcUser == "" || bmcPass == "" {
-		return ErrBMCCredentials
+		return cleanup, ErrBMCCredentials
 	}
 
 	addr, err := netip.ParseAddr(bmcAddr)
 	if err != nil {
-		return err
+		return cleanup, err
 	}
 
 	// Add server
 	server := sservice.Server{UUID: serverID, Name: serverID.String(), FacilityCode: facilityCode}
 	_, _, err = s.client.Create(otelCtx, server)
 	if err != nil {
-		return err
+		return cleanup, err
 	}
+	createStatus.serverCreated = true
 
 	// Add server BMC credential
-	_, err = s.client.SetCredential(otelCtx, serverID, "bmc", bmcUser, bmcPass)
+	_, err = s.client.SetCredential(otelCtx, serverID, secretSlug, bmcUser, bmcPass)
 	if err != nil {
-		return err
+		return cleanup, err
 	}
+	createStatus.credentialCreated = true
 
 	// Add server BMC IP attribute
 	addrAttr := fmt.Sprintf(`{"address": %q}`, addr.String())
 	bmcIPAttr := sservice.Attributes{Namespace: rservice.ServerAttributeNSBmcAddress, Data: []byte(addrAttr)}
 	_, err = s.client.CreateAttributes(otelCtx, serverID, bmcIPAttr)
-	return err
+	if err != nil {
+		return cleanup, err
+	}
+	createStatus.attributesCreated = true
+	return cleanup, nil
 }
 
 // GetServer returns the facility for the requested server id.
--- a/pkg/api/v1/routes/handlers.go
+++ b/pkg/api/v1/routes/handlers.go
@@ -161,34 +161,41 @@ func (r *Routes) serverEnroll(c *gin.Context) (int, *v1types.ServerResponse) {
 	}
 
 	// Creates a server record in FleetDB.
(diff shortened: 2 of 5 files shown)


internal/fleetdb/fleetdb.go
```diff
@@ -15,45 +15,86 @@ import (
...
 	_, _, err = s.client.Create(ctx, server)
 	if err != nil {
-		return err
+		return cleanup, err
 	}
+	createStatus.serverCreated = true
 
 	// Add server BMC credential
-	_, err = s.client.SetCredential(ctx, serverID, "bmc", bmcUser, bmcPass)
+	_, err = s.client.SetCredential(ctx, serverID, secretSlug, bmcUser, bmcPass)
```
DoctorVin, line 85: `gomnd` strikes again. 😒

pkg/api/v1/routes/handlers.go
```diff
@@ -212,8 +212,21 @@ func (r *Routes) serverEnroll(c *gin.Context) (int, *v1types.ServerResponse) {
...
+		if err := cleanup(); err != nil {
+			r.logger.WithFields(logrus.Fields{
+				"error": err,
+			}).Error("failed to cleanup server")
+		}
+	}()
 	// Creates a server record in FleetDB.
-	if err = r.fleetDBClient.AddServer(c.Request.Context(), serverID, params.Facility, params.IP, params.Username, params.Password); err != nil {
+	if cleanup, err = r.fleetDBClient.AddServer(c.Request.Context(), serverID, params.Facility, params.IP, params.Username, params.Password); err != nil {
+		shouldCleanup = true
```
DoctorVin, line 229: Why not clean up the database record when we discover the error? That is, why push this to the handler instead of doing it at the FleetDB client layer?

pkg/api/v1/routes/handlers.go
```diff
@@ -227,20 +240,26 @@ func (r *Routes) serverEnroll(c *gin.Context) (int, *v1types.ServerResponse) {
 	// Delete params related to add servers.
 	var inventoryArgs rctypes.InventoryTaskParameters
```
DoctorVin, line 241: We shouldn't be accepting this from consumers of the API, we should just dictate it as `outofband`, `collect_bios_cfg = true`, `collect_firmware_true`. We already have `asset_id` too. That save a lot of boilerplate concerns over cleaning up on JSON marshaling that in practice will never be a concern.

@joelrebel would you agree that these are the correct inventory parameters to pass here?

pkg/api/v1/routes/handlers.go
```diff
@@ -227,20 +240,26 @@ func (r *Routes) serverEnroll(c *gin.Context) (int, *v1types.ServerResponse) {
...
 			Message: err.Error(),
 		}
 	}
 	conditionCreate.Parameters = inventoryParams
 	newCondition := conditionCreate.NewCondition(rctypes.Inventory)
 
-	return r.conditionCreate(otelCtx, newCondition, serverID, params.Facility)
+	st, resp := r.conditionCreate(otelCtx, newCondition, serverID, params.Facility)
+	if st != http.StatusOK {
+		shouldCleanup = true
```
DoctorVin, line 260: By eliminating the marshaling stuff, we can call `cleanup` here directly instead of `deferring` it.

DoctorVin (changes requested):
We're going to collide on `fleetdb.go` I think. Some minor comments about how we handle the parameters for an inventory.

pkg/api/v1/routes/handlers.go
Alva8756: Do you suggest we hard code the parameters for creating the new condition?

This means we only need to set `asset_id` to `serverID` where other parameters will be hard coded as `outofband`, `collect_bios_cfg = true`, `collect_firmware_true`?

pkg/api/v1/routes/handlers.go
Alva8756: `FleedDB`client layer doesn't know whether `conditionCreate()` is succeeded or not.
If users failed to create new condition, it looks like users may be able to re-create again? If we don't cleanup, users will get `duplicate` error when enroll server again.

I am not sure if my understanding is correct. I can move the cleanup into the `FleetDB` client layer if that's more appropriate.

DoctorVin: No, I was slow, but I understand what you're doing here, and you're right: if we fail to create or publish a condition, we need to roll the database change back. Consider a name change of `cleanup` to `rollback` maybe? That would make it clearer to me, but it might not be universal.

pkg/api/v1/routes/handlers.go
DoctorVin: Yes. I think clients of `serverEnroll` don't need to know (and shouldn't care) that we do an inventory as part of enrollment. It also insulates them from any changes we make later either in how we perform an inventory, or if we add additional steps to the enrollment process.

pkg/api/v1/routes/handlers.go
```diff
@@ -161,34 +161,41 @@ func (r *Routes) serverEnroll(c *gin.Context) (int, *v1types.ServerResponse) {
 	}
 
 	// Creates a server record in FleetDB.
-	if err = r.fleetDBClient.AddServer(c.Request.Context(), serverID, params.Facility, params.IP, params.Username, params.Password); err != nil {
+	rollback, err := r.fleetDBClient.AddServer(c.Request.Context(), serverID, params.Facility, params.IP, params.Username, params.Password)
```
DoctorVin, line 164: I think calling the cleanup function `rollback` here is a good move.

pkg/api/v1/routes/handlers.go
```diff
@@ -161,34 +161,41 @@ func (r *Routes) serverEnroll(c *gin.Context) (int, *v1types.ServerResponse) {
...
-		return http.StatusBadRequest, &v1types.ServerResponse{
-			Message: "invalid inventory params: " + err.Error(),
-		}
+	inventoryArgs := &rctypes.InventoryTaskParameters{
+		AssetID:               serverID,
+		Method:                rctypes.OutofbandInventory,
+		CollectBiosCfg:        true,
+		CollectFirwmareStatus: true,
 	}
 	inventoryParams, err := json.Marshal(inventoryArgs)
```
DoctorVin, line 183: A minor thing, but I think it helps a bit: because we control `inventoryArgs` directly, we could (should?) panic if the serialization fails. Basically it should *never* fail, and if it did, there's not anything a caller of this API could do about it. If we instead `panic` here, we get an alert for ConditionAPI restarting, we get a stack trace, and likely we get more, better feedback than waiting for some random inquiry from support asking about why this customer can't enroll their server.

I wouldn't block merging this PR if you didn't want to do it now.

pkg/api/v1/routes/handlers_test.go
```diff
@@ -84,9 +84,7 @@ func asJSONBytes(t *testing.T, s *v1types.ServerResponse) []byte {
 	return b
 }
 
-// nolint:gocyclo // cyclomatic tests are cyclomatic
-func TestAddServer(t *testing.T) {
-	// mock repository
+func setupTestServer(t *testing.T) (*store.MockRepository, *fleetdb.MockFleetDB, *mockevents.MockStream, *gin.Engine, error) {
```
DoctorVin, line 87: :chefskiss:

pkg/api/v1/routes/handlers_test.go
```diff
@@ -306,6 +314,168 @@ func TestAddServer(t *testing.T) {
 	}
 }
 
+// nolint:gocyclo // cyclomatic tests are cyclomatic
+func TestAddServerRollback(t *testing.T) {
```
DoctorVin, line 318: It's definitely out-of-scope for this PR, but reading these tests convinces me that this is not the right layer for these tests, because we're testing handlers behavior by testing HTTP requests and responses instead of directly invoking the `Handlers` object.

DoctorVin (approved):
LGTM

pkg/api/v1/routes/handlers.go
Alva8756: Sent out #140

pkg/api/v1/routes/handlers_test.go
Alva8756: Can you please help provide more explanation?

Do you suggestion we mock [server service client](https://github.com/metal-toolbox/hollow-serverservice/blob/9089164ea19a188d3e5eb79d56892e23d9b2f121/pkg/api/v1/server_service.go#L28-L64) to get more test coverage?

PR #138, opened 2023-11-07
Vc/store refactor

DoctorVin:
#### What does this PR do
Remove the capability of using server-service as a condition store. Rework the repository interface a bit to simplify it and break some server-service details that leaked through it. Also move the server lookup functions to the fleetDB client interface so it's closer to its source-of-truth.


internal/store/nats.go
```diff
@@ -202,31 +174,23 @@ func (n *natsStore) List(ctx context.Context, srvID uuid.UUID, incState rctypes.
...
-		return nil, nil
+	var active *rctypes.Condition
+	for _, cond := range cr.Conditions {
+		if !rctypes.StateIsComplete(cond.State) {
+			le.WithFields(logrus.Fields{
+				"condition.id":    cond.ID,
+				"condition.state": string(cond.State),
+			}).Debug("found active condition for server")
+			active = cond
+			break
```
Alva8756, line 189: Curious we will only have 1 active condition at a time for given server-id right? (Maybe stupid questions just curious and want to learn)

Alva8756 (approved):

internal/store/nats.go
DoctorVin: Right, as things stand now, having an active condition blocks setting any other condition on a server. It's easier to do this than try and work out what conditions are compatible in a general way.

PR #140, opened 2023-11-10
Improve server_enroll error handling

Alva8756:
- Replace error response with panic to catch alert faster, as is suggested in https://github.com/metal-toolbox/conditionorc/pull/137/files#r1388602247
- Dump warning logs when fail to rollback server. It may be easier for us to query logs and tracking leaked servers.

### DIFF
--- a/internal/fleetdb/fleetdb.go
+++ b/internal/fleetdb/fleetdb.go
@@ -57,9 +57,7 @@ func (s *fleetDBImpl) AddServer(ctx context.Context, serverID uuid.UUID, facilit
 			server := sservice.Server{UUID: serverID, Name: serverID.String(), FacilityCode: facilityCode}
 			_, err := s.client.Delete(ctx, server)
 			if err != nil {
-				s.logger.WithFields(logrus.Fields{
-					"error": err,
-				}).Warningf("server enroll failed to rollback server")
+				s.logger.WithError(err).Warning("server enroll failed to rollback server")
 				return err
 			}
 		}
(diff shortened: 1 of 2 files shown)


internal/fleetdb/fleetdb.go
```diff
@@ -57,20 +57,29 @@ func (s *fleetDBImpl) AddServer(ctx context.Context, serverID uuid.UUID, facilit
 			server := sservice.Server{UUID: serverID, Name: serverID.String(), FacilityCode: facilityCode}
 			_, err := s.client.Delete(ctx, server)
 			if err != nil {
+				s.logger.WithFields(logrus.Fields{
+					"error": err,
+				}).Warningf("server enroll failed to rollback server")
```
coffeefreak101, line 62: nit: It might be cleaner to simply use `WithError()`:
```
s.logger.WithError(err).Warning("server enroll failed to rollback server")
```

DoctorVin: +1 Plus you'll be more consistent with our usage of `WithError` elsewhere in the codebase.

pkg/api/v1/routes/handlers.go
```diff
@@ -182,10 +182,8 @@ func (r *Routes) serverEnroll(c *gin.Context) (int, *v1types.ServerResponse) {
 	}
 	inventoryParams, err := json.Marshal(inventoryArgs)
 	if err != nil {
-		rollbackErr := rollback()
-		return http.StatusBadRequest, &v1types.ServerResponse{
-			Message: err.Error() + fmt.Sprintf("server rollback err: %v", rollbackErr),
-		}
+		_ = rollback()
```
DoctorVin, line 185: A log message here would be helpful.

DoctorVin (dismissed):

DoctorVin (approved):

PR #146, opened 2023-11-15
Support Server Delete

Alva8756:
Support `DeleteServer`.

New API: `/api/v1/serverDelete/{$uuid}`

**Tested in sandbox**
Existing Server
```
                 id                  |                 name                 | facility_code |          created_at           |          updated_at           |          deleted_at
---------------------------------------+--------------------------------------+---------------+-------------------------------+-------------------------------+--------------------------------
b3744970-47a0-4261-abd5-de19d7f7c6bd | b3744970-47a0-4261-abd5-de19d7f7c6bd | sandbox       | 2023-11-17 04:49:52.082456+00 | 2023-11-17 04:49:52.082456+00 | NULL
```

First Delete
```
$ curl -i -X DELETE 'localhost:9001/api/v1/serverDelete/b3744970-47a0-4261-abd5-de19d7f7c6bd'
HTTP/1.1 200 OK
...
{"message":"server detele","records":{"serverID":"b3744970-47a0-4261-abd5-de19d7f7c6bd"}}
```
```
                 id                  |                 name                 | facility_code |          created_at           |          updated_at           |          deleted_at
---------------------------------------+--------------------------------------+---------------+-------------------------------+-------------------------------+--------------------------------
b3744970-47a0-4261-abd5-de19d7f7c6bd | b3744970-47a0-4261-abd5-de19d7f7c6bd | sandbox       | 2023-11-17 04:49:52.082456+00 | 2023-11-17 04:49:52.082456+00 | 2023-11-26 22:36:12.538992+00
```

Second Delete
```
$ curl -i -X DELETE 'localhost:9001/api/v1/serverDelete/b3744970-47a0-4261-abd5-de19d7f7c6bd'
HTTP/1.1 500 Internal Server Error
...
{"message":"hollow client received a server error - response code: 404, message: resource not found, details: sql: no rows in result set"}
```

Discuss:
For the second delete, we can return `404` to match server service error message.
Or use `500` referring the error comes from the server.
I don't have preference about it.

### DIFF
--- a/pkg/api/v1/routes/handlers.go
+++ b/pkg/api/v1/routes/handlers.go
@@ -98,7 +98,7 @@ func (r *Routes) serverConditionCreate(c *gin.Context) (int, *v1types.ServerResp
 	facilityCode, err := r.serverFacilityCode(otelCtx, serverID)
 	if err != nil {
 		return http.StatusInternalServerError, &v1types.ServerResponse{
-			Message: "error looking up facility on non-existence server: " + err.Error(),
+			Message: "facility lookup error: " + err.Error(),
 		}
 	}
 
@@ -155,8 +155,10 @@ func (r *Routes) serverDelete(c *gin.Context) (int, *v1types.ServerResponse) {
 	}
 
 	if err := r.fleetDBClient.DeleteServer(c.Request.Context(), serverID); err != nil {
-		return http.StatusInternalServerError, &v1types.ServerResponse{
-			Message: err.Error(),
+		if !strings.Contains(err.Error(), "404") {
+			return http.StatusInternalServerError, &v1types.ServerResponse{
+				Message: err.Error(),
+			}
 		}
 	}
 
--- a/pkg/api/v1/routes/handlers.go
+++ b/pkg/api/v1/routes/handlers.go
@@ -98,7 +98,7 @@ func (r *Routes) serverConditionCreate(c *gin.Context) (int, *v1types.ServerResp
 	facilityCode, err := r.serverFacilityCode(otelCtx, serverID)
 	if err != nil {
 		return http.StatusInternalServerError, &v1types.ServerResponse{
-			Message: err.Error(),
+			Message: "error looking up facility on non-existence server: " + err.Error(),
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/api/v1/routes/handlers.go
```diff
@@ -119,6 +119,34 @@ func (r *Routes) serverConditionCreate(c *gin.Context) (int, *v1types.ServerResp
...
+	serverID, _ := uuid.Parse(c.Param("uuid"))
+	if err := r.fleetDBClient.DeleteServer(c.Request.Context(), serverID); err != nil {
+		return http.StatusInternalServerError, &v1types.ServerResponse{
+			Message: err.Error(),
+		}
+	}
+
+	// Do we need to
+	// publish/update condtion letting subscriber knows that this condition is no longer valid?
+	// or is repository.Delete already good enough?
```
DoctorVin, line 136: I would not touch any conditions that are in flight, because it will cause more problems than it solves. If someone deletes a server while a condition is being worked on, the fleetDB updates will fail, and that will probably cause the executing condition to fail as well. That's acceptable, in my view.

You could use `r.repository.GetActiveCondition` to make sure that the server was actually not being worked on. I think that's probably a better test. If there is an active condition on the server, fail the delete.

pkg/api/v1/routes/handlers.go
```diff
@@ -119,6 +119,34 @@ func (r *Routes) serverConditionCreate(c *gin.Context) (int, *v1types.ServerResp
...
+
+	kind := rctypes.Kind(c.Param("conditionKind"))
+	if err := r.repository.Delete(otelCtx, serverID, kind); err != nil {
```
DoctorVin, line 139: I have a PR that removes the `Delete` method from the Repository interface.

DoctorVin (commented):
A decent start so far. Let me know if you have questions.

pkg/api/v1/routes/handlers.go
```diff
@@ -119,6 +119,53 @@ func (r *Routes) serverConditionCreate(c *gin.Context) (int, *v1types.ServerResp
...
+		r.logger.WithFields(logrus.Fields{
+			"serverID": id,
+		}).Info("bad serverID")
+
+		return http.StatusBadRequest, &v1types.ServerResponse{
+			Message: err.Error(),
+		}
+	}
+
+	active, err := r.repository.GetActiveCondition(otelCtx, serverID)
```
DoctorVin, line 145: 👍

DoctorVin (changes requested):
3 minor things, ready to stamp when updated.

DoctorVin:
We also might consider removing the configuration gate on serverEnroll now that we're happy with it, but that should be a separate PR.

DoctorVin (approved):

PR #147, opened 2023-11-16
[FS-1045]Create a new server ID when no uuid param found

Alva8756:


pkg/api/v1/routes/handlers.go
```diff
@@ -125,17 +125,6 @@ func (r *Routes) serverEnroll(c *gin.Context) (int, *v1types.ServerResponse) {
 	span.SetAttributes(attribute.KeyValue{Key: "serverId", Value: attribute.StringValue(id)})
```
DoctorVin, line 125: Probably should move this down to around line 151, so we don't have multiple attributes for `server id` on this span.

DoctorVin (approved):
LGTM

DoctorVin (approved):

PR #149, opened 2023-11-20
Add a client API for enrolling server

Alva8756:
Tested locally: `mctl` -> `sandbox`

It looks `client_test.go` is broken currently. I didn't find an easy way to fix it.

### DIFF
--- a/pkg/api/v1/client/client.go
+++ b/pkg/api/v1/client/client.go
@@ -83,3 +83,10 @@ func (c *Client) ServerConditionCreate(ctx context.Context, serverID uuid.UUID, 
 
 	return c.post(ctx, path, conditionCreate)
 }
+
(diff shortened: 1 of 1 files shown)


pkg/api/v1/client/client.go
```diff
@@ -83,3 +83,10 @@ func (c *Client) ServerConditionCreate(ctx context.Context, serverID uuid.UUID,
 
 	return c.post(ctx, path, conditionCreate)
 }
+
+func (c *Client) ServerEnroll(ctx context.Context, serverID string, conditionCreate v1types.ConditionCreate) (*v1types.ServerResponse, error) {
```
DoctorVin, line 87: Question for thought: would it be easier/better to break this out into a method with more "primitive" arguments (i.e. strings or golang std library types)?

DoctorVin (approved):
Fine as is. Maybe longer-term we think about changing up the client code a bit.

PR #153, opened 2023-11-28
Vc/condition status endpoint

DoctorVin:
#### What does this PR do
Add an endpoint that is more aware of work having multiple conditions. This PR redoes the `serverConditionGet` endpoint as `conditionStatus` and uses newer Repository functions to query the data.


DoctorVin:
I have no idea why the one lint task fails and the others succeed. I can't replicate this in my local environment and the lint check passes.

pkg/api/v1/routes/routes.go
```diff
@@ -149,14 +149,14 @@ func (r *Routes) Routes(g *gin.RouterGroup) {
 		servers.POST("/firmwareInstall", r.composeAuthHandler(createScopes("condition")),
 			wrapAPICall(r.firmwareInstall))
 
+		// Generalized API for any condition status (for cases where some server work
+		// has multiple conditions involved and the caller doesn't know what they might be)
+		servers.GET("/conditionStatus", r.composeAuthHandler(readScopes("condition")),
+			wrapAPICall(r.conditionStatus))
```
coffeefreak101, line 155: Not sure how much we care about this, but I think "/conditions" would be a more RESTful way to express an endpoint that returns a collection of conditions.

DoctorVin: That's a good point, and I do like being super RESTful; my only misgiving with that is that this isn't a general purpose condition history endpoint. It's intended as more of "what's going on with this server now (or in the immediate past)."

That said, if we get a consensus for change, I'm not opposed to it either.

joelrebel: > It's intended as more of "what's going on with this server now (or in the immediate past)."

nit pick: I think this could be just `/status`, so we drop the `condition` prefix.

joelrebel:
mocks needs updating https://github.com/metal-toolbox/conditionorc/actions/runs/7024302404/job/19112726598?pr=153#step:4:35

DoctorVin:
> mocks needs updating https://github.com/metal-toolbox/conditionorc/actions/runs/7024302404/job/19112726598?pr=153#step:4:35

That reflects the state of a single commit. If the mocks were broken, the tests would not run and all build/lint steps would fail.

PR #155, opened 2023-11-29
[FS-1032 Minor]Remove duplicate delete

Alva8756:
#### What does this PR do
This change is according to the thread there 
https://equinix.slack.com/archives/C05U5L6D54Z/p1701254188201809?thread_ts=1701111416.845089&cid=C05U5L6D54Z

### DIFF
--- a/pkg/api/v1/routes/handlers_test.go
+++ b/pkg/api/v1/routes/handlers_test.go
@@ -423,7 +423,7 @@ func TestDeleteServer(t *testing.T) {
 					Times(1)
 			},
 			func(t *testing.T) *http.Request {
-				url := fmt.Sprintf("/api/v1/server/%v", mockServerID)
+				url := fmt.Sprintf("/api/v1/servers/%v", mockServerID)
 				request, err := http.NewRequestWithContext(context.TODO(), http.MethodDelete, url, nil)
 				if err != nil {
 					t.Fatal(err)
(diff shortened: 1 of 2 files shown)


pkg/api/v1/routes/routes.go
```diff
@@ -139,10 +139,10 @@ func (r *Routes) Routes(g *gin.RouterGroup) {
 	if r.enableServerEnroll {
 		serverEnroll := g.Group("/serverEnroll")
 		serverEnroll.POST("/:uuid", r.composeAuthHandler(createScopes("server")), wrapAPICall(r.serverEnroll))
-		serverDelete := g.Group("/serverDelete")
-		serverDelete.DELETE("/:uuid", r.composeAuthHandler(createScopes("server")), wrapAPICall(r.serverDelete))
 		// Create a new server ID when uuid is not provided.
 		serverEnroll.POST("/", r.composeAuthHandler(createScopes("server")), wrapAPICall(r.serverEnroll))
+		serverDelete := g.Group("/server")
```
DoctorVin, line 144: we should use the servers group at line 148 for this. I think you can just move that line above this block. Then line 144 is not needed and line 145 becomes: `servers.DELETE("", r.composeAuthHandler...`

pkg/api/v1/routes/handlers_test.go
```diff
@@ -423,7 +423,7 @@ func TestDeleteServer(t *testing.T) {
 					Times(1)
 			},
 			func(t *testing.T) *http.Request {
-				url := fmt.Sprintf("/api/v1/serverDelete/%v", mockServerID)
+				url := fmt.Sprintf("/api/v1/server/%v", mockServerID)
```
DoctorVin, line 426: The target url should be `api/v1/servers/{serverID}`. Note `servers` plural, not `server` singular.

DoctorVin (changes requested):

DoctorVin (approved):
LGTM

PR #172, opened 2023-12-19
Update to rivets v0.2.2 which renames the kv package

joelrebel:


DoctorVin (changes requested):
LGTM, just get rid of the duplicate import.

DoctorVin (approved):
Go on green.

PR #181, opened 2024-01-29
refresh the status kv watcher on more than a single nil on the updates channel

DoctorVin:
This change tries to capture and correct error conditions when the updates channel from a NATS KV watcher gets closed.

### DIFF
--- a/internal/orchestrator/updates.go
+++ b/internal/orchestrator/updates.go
@@ -123,6 +123,9 @@ func (o *Orchestrator) startConditionWatchers(ctx context.Context,
 
 		go func() {
 			defer wg.Done()
+			// NATS will send an nil if the connection is live but there are no updates. We expect one per use
+			// of the update channel, sent before it transitions to a blocking behavior.
+			sawNil := false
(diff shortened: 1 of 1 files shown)


internal/orchestrator/updates.go
```diff
@@ -132,7 +135,21 @@ func (o *Orchestrator) startConditionWatchers(ctx context.Context,
 					watcher.Stop()
 				case entry := <-watcher.Updates():
 					if entry == nil {
+						if sawNil {
```
joelrebel, line 138: can this be covered by a test, by any chance?

DoctorVin: Can't cover this by a test, we'd have to engineer a broken connection to NATS, which is not straightforward. In general it is hard to test error handling at the NATS interface because engineering the correct fault is not possible.

I feel like moving this into a capture in this block actually detracts from readability.

joelrebel: I tend to prefer smaller blocks, but this isn't a blocker so get this merged.

joelrebel (approved):

PR #185, opened 2024-02-26
[FS-1253] make 'no active condition' handling uniform

DoctorVin:
#### What does this PR do

We observed a panic in the condition update routine under load where a delayed or replayed update resulted in a nil, no-error return from the `GetActiveCondition` call. Further investigation found the semantics of this call were inconsistent, with some code checking for `nil` explicitly and some assuming that `ErrConditionNotFound` would be returned.

Here we change this to be more-consistent, where we explicitly return `ErrConditionNotFound` when such is appropriate (that is, either there is no condition to find, *or* that condition is in a completed state).

### DIFF
--- a/internal/orchestrator/updates.go
+++ b/internal/orchestrator/updates.go
@@ -11,6 +11,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/metal-toolbox/conditionorc/internal/metrics"
 	"github.com/metal-toolbox/conditionorc/internal/status"
+	"github.com/metal-toolbox/conditionorc/internal/store"
 	"github.com/nats-io/nats.go"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
@@ -324,6 +325,11 @@ func (o *Orchestrator) eventUpdate(ctx context.Context, evt *v1types.ConditionUp
 
 		// queue any follow-on work as required
 		active, err := o.repository.GetActiveCondition(ctx, evt.ConditionUpdate.ServerID)
+		if err != nil && errors.Is(err, store.ErrConditionNotFound) {
+			// nothing more to do
+			return nil
+		}
+
 		if err != nil {
 			o.logger.WithError(err).WithFields(logrus.Fields{
 				"condition.id": evt.ConditionUpdate.ConditionID,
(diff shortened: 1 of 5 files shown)


pkg/api/v1/events/handlers.go
```diff
@@ -86,8 +86,9 @@ func (h *Handler) UpdateCondition(ctx context.Context, updEvt *v1types.Condition
 	if err != nil {
 		if errors.Is(err, store.ErrConditionNotFound) {
 			h.logger.WithFields(logrus.Fields{
-				"serverID":      updEvt.ConditionUpdate.ServerID,
-				"conditionKind": updEvt.Kind,
+				"server_id":      updEvt.ConditionUpdate.ServerID,
```
joelrebel, line 89: can this be done throughout the code base so its consistent - `s/\"serverID\"/\"server_id\"/g` and `s/\"conditionID\"/\"condition_id\"/g`?

internal/store/nats.go
```diff
@@ -115,16 +115,12 @@ func (n *natsStore) GetActiveCondition(ctx context.Context, serverID uuid.UUID)
...
-		return nil, nil
-	}
-
 	if err != nil {
 		return nil, err
 	}
 
 	if rctypes.StateIsComplete(cr.State) {
-		return nil, nil
+		return nil, ErrConditionNotFound
```
joelrebel, line 123: why does this return `ErrConditionNotFound`, when clearly it does exist - or should this error be instead named 'ErrNoActiveConditionFound'?

internal/orchestrator/updates.go
```diff
@@ -324,6 +325,11 @@ func (o *Orchestrator) eventUpdate(ctx context.Context, evt *v1types.ConditionUp
 
 		// queue any follow-on work as required
 		active, err := o.repository.GetActiveCondition(ctx, evt.ConditionUpdate.ServerID)
+		if err != nil && errors.Is(err, store.ErrConditionNotFound) {
```
joelrebel, line 328: Its a bit misleading when I read this - should we have two error types here,
`ErrConditionNotFound`, `ErrNoActiveCondition` so they can be explicit

pkg/api/v1/events/handlers.go
DoctorVin: We can but not in this pull-request which I want to keep targeted to the issue at hand.

internal/orchestrator/updates.go
DoctorVin: While I do value explicitness here, in the end `ErrConditionNotFound` (which arguably should be `ErrNoActiveConditionFound` but... ) wind up being treated exactly the same. It's an error if you were expecting an active condition and not if you weren't. As far as the name goes, I'm ambivalent. We can change it if you think it's useful but I don't see it as adding a bunch of friction to further development or maintenance.

joelrebel: fair, it can stay unless it comes up again

joelrebel (approved):

PR #195, opened 2024-03-18
Switch to fleetdb, Go 1.22 and rivets v1.0.1

joelrebel:
#### What does this PR do

Also fixes up dependency updates reported by renovate.


internal/app/config.go
```diff
@@ -44,10 +44,10 @@ type Configuration struct {
 	// ConditionDefinitions holds one or more condition definitions the conditionorc API, orchestrator support.
 	ConditionDefinitions rctypes.Definitions `mapstructure:"conditions"`
 
-	// ServerserviceOptions defines the serverservice client configuration parameters
+	// FleetDBAPIOptions defines the serverservice client configuration parameters
 	//
 	// This parameter is required when StoreKind is set to serverservice.
-	ServerserviceOptions ServerserviceOptions `mapstructure:"serverservice"`
+	FleetDBAPIOptions FleetDBAPIOptions `mapstructure:"serverservice"`
```
DoctorVin, line 50: change the `mapstructure` tag? Or capture a task to do so later?

internal/app/config.go
```diff
@@ -241,61 +241,61 @@ func (a *App) envVarNatsOverrides() error {
...
+	endpointURL, err := url.Parse(a.Config.FleetDBAPIOptions.Endpoint)
 	if err != nil {
 		return errors.New("serverservice endpoint URL error: " + err.Error())
 	}
 
-	a.Config.ServerserviceOptions.EndpointURL = endpointURL
+	a.Config.FleetDBAPIOptions.EndpointURL = endpointURL
 
 	if a.v.GetString("serverservice.disable.oauth") != "" {
-		a.Config.ServerserviceOptions.DisableOAuth = a.v.GetBool("serverservice.disable.oauth")
```
DoctorVin, line 255: should change the names on the environment variables too

DoctorVin (changes requested):
Would be nicer to gang-in the complete swap from `serverservice` to `fleetdb` even in the config files and environment variables.

internal/app/config.go
joelrebel: Yep, I intend to take this on separately - since it involves changing the k8s templates, and that can be done in a separate sweep across the apps.

internal/app/config.go
joelrebel: yep, same as my previous reply

DoctorVin (approved):

PR #197, opened 2024-03-27
server: fail to start with OIDC if configuration is not defined

joelrebel:


jakeschuurmans (dismissed):

internal/server/server.go
DoctorVin, line 1: The changes here are not required, as the zero-value of a slice is nil in Golang: cf. https://go.dev/tour/moretypes/12

DoctorVin (changes requested):

DoctorVin (approved):

PR #200, opened 2024-04-23
Orchestrator reconcile

joelrebel:
#### What does this PR do

- Adds a few metrics
- A few method renames, refactors
- Enables reconcile loop based on Vince's work
- Purges old code

depends on  https://github.com/metal-toolbox/rivets/pull/51
To be merged after https://github.com/metal-toolbox/conditionorc/pull/203


internal/orchestrator/updates.go
```diff
@@ -302,67 +328,87 @@ func (o *Orchestrator) eventUpdate(ctx context.Context, evt *v1types.ConditionUp
...
+	).Inc()
+
+	// queue any follow-on work as required
+	return o.queueFollowingCondition(ctx, evt)
+}
+
+// Queue up follow on conditions
+//
+// TODO: Q. how do we know the follow on work is ordered as expected, or is the assumption that there could be only one other condition queued for this server ?
+// TODO: Q. if the current condition has failed, we most likely don't want to queue the next condition?
```
DoctorVin, line 364: If a condition has failed, when we update the store (cf. line 327) we will mark the entire `conditionRecord` as failed. When we call `o.repository.GetActiveCondition` that will then return `ErrConditionNotFound`, signifying that there is nothing more to do, and we short-circuit at line 369.

That logic for `GetActiveCondition` is [here](https://github.com/metal-toolbox/conditionorc/blob/16100161c40dc89f0543305a4069cca5aeaa5eeb/internal/store/nats.go#L109) while the update is [here](https://github.com/metal-toolbox/conditionorc/blob/16100161c40dc89f0543305a4069cca5aeaa5eeb/internal/store/nats.go#L227) and the failure case is explicitly [tested](https://github.com/metal-toolbox/conditionorc/blob/16100161c40dc89f0543305a4069cca5aeaa5eeb/internal/store/nats_test.go#L231)

internal/orchestrator/updates.go
```diff
@@ -388,28 +434,34 @@ func (o *Orchestrator) eventNeedsReconciliation(evt *v1types.ConditionUpdateEven
 	})
 
 	lastTime, err := registry.LastContact(evt.ControllerID)
-	if errors.Is(err, nats.ErrKeyNotFound) {
-		return true
-	}
+	if err != nil {
```
DoctorVin, line 437: you don't need this. `nil` is not `nats.ErrKeyNotFound`.

_Edit:_ I'm ambivalent at best on this particular refactor. I don't think it adds anything, but instead just pushes code around in different ways.

internal/orchestrator/updates.go
```diff
@@ -388,28 +434,34 @@ func (o *Orchestrator) eventNeedsReconciliation(evt *v1types.ConditionUpdateEven
...
+	metrics.ConditionReconcileStale.With(
+		prometheus.Labels{"conditionKind": string(evt.Kind)},
+	).Inc()
+
 	return false
 }
 
-func (o *Orchestrator) startReconciler(ctx context.Context, wg *sync.WaitGroup) {
-	wg.Add(1)
+func (o *Orchestrator) startReconciler(ctx context.Context) {
```
DoctorVin, line 460: Don't do this. I explicitly use a different `WaitGroup` on process start to handle this. Ganging it all into the member `WaitGroup` isn't needed.

pkg/api/v1/events/handlers.go
DoctorVin, line 1: 🎉

DoctorVin (changes requested):
Let's chat about your refactors in `internal/orchestrator/updates.go`.

internal/orchestrator/updates.go
joelrebel: why does it have to be separate waitgroups instead of the shared, is it better practice?

internal/orchestrator/updates.go
joelrebel: Thanks for explaining that, the linked test helped answer my questions as well.

internal/metrics/metrics.go
```diff
@@ -74,6 +78,30 @@ func init() {
 		},
 		[]string{"conditionKind"},
 	)
+
+	ConditionInKV = promauto.NewCounterVec(
```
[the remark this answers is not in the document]
joelrebel: The plan here is to collect metrics on the conditions in the `active-conditions` KV.

You're right this is ideally a Gauge, although, as a Gauge - the collection code gets more involved - since the Gauge value has to be set to the total for each Condition  grouped by its `conditionKind` and `state` values.

With the Counter type instead we just increment the value as we're iterating through and we use [delta](https://prometheus.io/docs/prometheus/latest/querying/functions/#delta) in the visualisation to obtain the difference over time. Let me know if Im missing something here.

internal/orchestrator/updates.go
joelrebel: Are you saying the change is incorrect here?

The point with the change was to group the error checks and return earlier, since that is easier to grok (imo),
as part of that it un-indents the additional else clause.

old
```go
	lastTime, err := registry.LastContact(evt.ControllerID)
	if errors.Is(err, nats.ErrKeyNotFound) {
		return true
	}

	if err == nil {
		since := time.Since(lastTime)
		le.WithFields(logrus.Fields{
			"last.update":    evt.UpdatedAt.String(),
			"worker.checkin": since.String(),
		}).Debug("long running event caught by reconciler")
	} else {
		le.WithError(err).Warn("error checking registry")
	}

	return false
```

new
```go
	lastTime, err := registry.LastContact(evt.ControllerID)
	if err != nil {
		if errors.Is(err, nats.ErrKeyNotFound) {
			return true
		}

		le.WithError(err).Warn("error checking registry")
		return false
	}

	// we don't attempt to clean up, but collect metrics on it
	since := time.Since(lastTime)
	le.WithFields(logrus.Fields{
		"last.update":    evt.UpdatedAt.String(),
		"worker.checkin": since.String(),
	}).Debug("long running event caught by reconciler")

	metrics.ConditionReconcileStale.With(
		prometheus.Labels{"conditionKind": string(evt.Kind)},
	).Inc()

	return false
```

internal/metrics/metrics.go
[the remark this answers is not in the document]
DoctorVin: the composites tracked by `active-conditions` don't have a `kind` per se. 

I think if you want to track active conditions, we're probably better served by having a small go-routine that enumerates the KV and sets the gauge on a cadence. My experience in using PromQL counters for a gauge is that the queries get weird and brittle quickly. We should keep this metric more intuitive.

internal/orchestrator/updates.go
```diff
@@ -77,7 +78,7 @@ func (o *Orchestrator) kvStatusPublisher(ctx context.Context) {
 			o.logger.Debug("stopping KV update listener")
 			stop = true
 
-		// retrieve and process events sent by controllers.
+		// retrieve and process events sent by controllers KV updates.
```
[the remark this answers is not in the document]
DoctorVin: `s/events/updates` IMO.
I tend to refer to "things sent on a Jetstream subject" as "messages". Using messages and updates avoids the overloaded `event` term.

internal/orchestrator/updates.go
DoctorVin: The member WaitGroup could be used by different callers, for different purposes, but they share a `Wait()` because they are the same object.

internal/orchestrator/updates.go
DoctorVin: I think we're just arguing code aesthetics, and that's the root of the issue. This change doesn't make meaningful changes aside shifting things about, and I don't find the new organization an improvement.

internal/metrics/metrics.go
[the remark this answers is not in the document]
joelrebel: thats a fair point, I'll leave a note in the code and have it taken up separately

internal/orchestrator/updates.go
joelrebel: Clarity in code layout is an improvement. 
The proposal here to have a single block within which all error cases handled, which to me is clearer. Given that we're maintaining this code together, I don't see why we should not proceed with this change.

internal/orchestrator/updates.go
joelrebel: Yep thats a fair point. 
Although in this case the reconciler routine is being spawned by the Orchestrator [directly](https://github.com/metal-toolbox/conditionorc/blob/f64bd86ecd3f4a312f1a320e399d92822e1c20b3/internal/orchestrator/orchestrator.go#L137) so it would be ideal to have it use the parent waitgroup.. or how do you propose it be done in this case?

internal/orchestrator/updates.go
DoctorVin: I 100% do not agree that this is "clearer."

internal/orchestrator/updates.go
DoctorVin: That imposes a burden on a maintainer to be sure that the member WaitGroup isn't actually being used for something else during system start, and it's vulnerable to someone making a change that compiles but leads to odd and inconsistent behavior at runtime that will be difficult to debug.

DoctorVin (approved):

internal/orchestrator/updates.go
```diff
@@ -263,37 +291,190 @@ func eventUpdateFromKV(ctx context.Context, kve nats.KeyValueEntry,
...
+			}
+
+			updateEvts[evt.ServerID.String()] = evt
+		}
+	}
+
+	return filterToReconcile(filteredRecords, updateEvts)
+}
+
+func (o *Orchestrator) getEventsToReconcile(ctx context.Context) (evts []*v1types.ConditionUpdateEvent) {
```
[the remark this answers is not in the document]
joelrebel: In this case `getEventsToReconcile()` is limited to the configured facility. I've added tests cases for `activeConditionsToReconcile` to validate the reconcile is limited to the configured facility. Thanks for the suggestion.

DoctorVin (approved):

PR #203, opened 2024-05-03
Active condition records reconcile

joelrebel:
#### What does this PR do

These changes ensures that,

1. The `active-conditions` KV record will be updated for cases where a Condition is not listed in the `$facility.$condition` KV because the Controller isn't running or is malfunctioning such that it cannot publish a status.
2. The `active-condition` record is missing and the Condition's needs to be re-created in the active-conditions.

### DIFF
--- a/internal/orchestrator/updates.go
+++ b/internal/orchestrator/updates.go
@@ -674,11 +674,26 @@ func (o *Orchestrator) reconcileActiveConditionRecords(ctx context.Context) {
 			"kind":           string(cond.Kind),
 		})
 
-		if err := o.repository.Create(ctx, cond.Target, cond); err != nil {
-			le.WithError(err).Warn("reconciler condition record create")
+		_, err := o.repository.Get(ctx, cond.Target)
+		if err != nil {
+			// create record if it doesn't exist
+			if errors.Is(err, store.ErrConditionNotFound) {
+				if errCreate := o.repository.Create(ctx, cond.Target, cond); errCreate != nil {
+					le.WithError(errCreate).Warn("reconciler condition record create")
+				}
+
+				le.Info("active-condition KV condition record reconciled - created new record")
+				continue
+			}
+
+			le.WithError(err).Warn("reconciler condition record get")
 			continue
 		}
 
-		le.Info("active-condition KV condition record reconciled")
+		// update record if it exists
+		if errUpdate := o.repository.Update(ctx, cond.Target, cond); errUpdate != nil {
+			le.WithError(errUpdate).Warn("reconciler condition record create")
+		}
+		le.Info("active-condition KV condition record reconciled - updated existing record")
 	}
 }
--- a/internal/orchestrator/updates.go
+++ b/internal/orchestrator/updates.go
@@ -284,11 +284,11 @@ func (o *Orchestrator) getEventsToReconcile(ctx context.Context) []*v1types.Cond
 	evts := []*v1types.ConditionUpdateEvent{}
 	for _, def := range o.conditionDefs {
 		kind := def.Kind
-
 		entries, err := status.GetAllConditions(kind, o.facility)
 		if err != nil {
 			o.logger.WithError(err).WithField("rctypes.kind", string(kind)).
-				Warn("reconciler error on condition lookup")
+				Warn("reconciler error in condition status lookup")
+			continue
 		}
 
 		for _, kve := range entries {
@@ -424,6 +424,7 @@ func (o *Orchestrator) queueFollowingCondition(ctx context.Context, evt *v1types
 	return nil
 }
 
+// This reconciles conditions in the Condition status KV - $condition.$facility
 func (o *Orchestrator) eventNeedsReconciliation(evt *v1types.ConditionUpdateEvent) bool {
 	// the last update should be later than the condition stale threshold
 	if time.Since(evt.ConditionUpdate.UpdatedAt) < rctypes.StaleThreshold {
@@ -462,7 +463,7 @@ func (o *Orchestrator) eventNeedsReconciliation(evt *v1types.ConditionUpdateEven
 		le.WithField(
 			"last.checkin",
 			time.Since(lastTime),
-		).Info("controller not checked in time exceeded")
+		).Info("controller check in timed out")
 		return true
 	}
 
@@ -485,6 +486,7 @@ func (o *Orchestrator) startReconciler(ctx context.Context, wg *sync.WaitGroup) 
 				keepRunning = false
 
 			case <-ticker.C:
+				// reconcile status KV entries
 				evts := o.getEventsToReconcile(ctx)
 				for _, evt := range evts {
 					le := o.logger.WithFields(logrus.Fields{
--- a/internal/orchestrator/updates.go
+++ b/internal/orchestrator/updates.go
@@ -293,7 +293,20 @@ func failedUpdateEventFromCondition(cond *rctypes.Condition) *v1types.ConditionU
 	}
 }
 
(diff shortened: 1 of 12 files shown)


internal/orchestrator/updates.go
```diff
@@ -279,16 +279,159 @@ func parseEventUpdateFromKV(ctx context.Context, kve nats.KeyValueEntry,
...
+	}
+
+	return creates, updates
+}
+
+func (o *Orchestrator) activeConditionsToReconcile(ctx context.Context) ([]*rctypes.Condition, []*v1types.ConditionUpdateEvent) {
+	// list condition records from the active-conditions KV
+	records, err := o.repository.List(ctx)
+	if err != nil {
+		o.logger.WithError(err).Error("condition record lookup error")
```
DoctorVin, line 372: an explicit return here would be clearer.

internal/orchestrator/updates.go
```diff
@@ -279,16 +279,159 @@ func parseEventUpdateFromKV(ctx context.Context, kve nats.KeyValueEntry,
...
+
+	// returned missing conditions from active-conditions KV to reconcile
+	creates := []*rctypes.Condition{}
+
+	// filter condition Kinds and records to be reconciled
+	filteredKinds, filteredRecords := filterIncompleteRecords(records)
+
+	// List Conditions in the Status KV and prepare an update payload
+	// for those to be reconciled.
+	for kind := range filteredKinds {
```
DoctorVin, line 386: We don't care what kind of condition it is that needs reconciliation. The criteria are based solely on time and controller liveness. I don't think _de facto_ batching does anything aside increase complexity.

internal/orchestrator/updates.go
```diff
@@ -279,16 +279,159 @@ func parseEventUpdateFromKV(ctx context.Context, kve nats.KeyValueEntry,
...
+
+		// parse status entries
+		for _, kve := range statusEntries {
+			evt, err := parseEventUpdateFromKV(ctx, kve, kind)
+			if err != nil {
+				o.logger.WithError(err).WithFields(logrus.Fields{
+					"rctypes.kind": string(kind),
+					"kv.key":       kve.Key(),
+				}).Warn("reconciler skipping malformed update")
+				continue
```
DoctorVin, line 406: We should fail a condition on a garbage update. Otherwise it will linger and accumulate (because the misbehaving controller is still doing things).

internal/orchestrator/updates.go
```diff
@@ -279,16 +279,159 @@ func parseEventUpdateFromKV(ctx context.Context, kve nats.KeyValueEntry,
...
+			}
+
+			if rctypes.StateIsComplete(evt.State) {
+				continue
+			}
+
+			updateEvts[evt.ServerID.String()] = evt
+		}
+
+		crts, updts := filterToReconcile(filteredRecords, updateEvts)
```
DoctorVin, line 416: by skipping the filtering for kind step, you can simply return the results of `filterToReconcile`

internal/orchestrator/updates.go
```diff
@@ -485,24 +629,69 @@ func (o *Orchestrator) startReconciler(ctx context.Context, wg *sync.WaitGroup)
...
+		le.Info("condition reconciled")
+
+		if err := o.notifier.Send(evt); err != nil {
+			le.WithError(err).Warn("reconciler event notification")
+		}
+	}
+}
+
+// reconcile active-condition KV entries for missing status KV entries and missing active-condition entries
+func (o *Orchestrator) reconcileActiveConditionRecords(ctx context.Context) {
```
DoctorVin, line 662: I feel like separating this process into `creates` and `updates` has the potential to miss things, like notifications. Why not conditionally create the `active_condition` record if the error from `UpdateCondition` is `ErrConditionNotFound`?

internal/store/nats.go
```diff
@@ -22,8 +22,8 @@ import (
 )
 
 var (
-	bucketName = "active-conditions"
-	bucketOpts = []kv.Option{
+	ActiveConditionBucket = "active-conditions"
```
DoctorVin, line 25: why is this bucket being exported from the module?

pkg/api/v1/types/types.go
```diff
@@ -152,6 +153,7 @@ func (c *ConditionUpdate) MergeExisting(existing *rctypes.Condition) (*rctypes.C
 		Parameters:            existing.Parameters,
 		State:                 c.State,
 		Status:                c.Status,
+		Target:                existing.Target,
```
DoctorVin, line 156: I feel like if we're going to guard against arbitrary bad data we should add a validation that `existing.Target` and `c.Target` should be the same

internal/orchestrator/updates.go
joelrebel: The notification was excluded on purpose here, since the `Notifier.Send()` expects a `ConditionUpdateEvent` and `creates` is a slice of `Condition` types, while `updates` is a slice of `ConditionUpdateEvents`.
 This is the main reason that they are individually handled as well - because they are different types and `UpdateCondition()` expects the `ConditionUpdateEvent` type.

internal/store/nats.go
joelrebel: I was working on some tests that are now pushed https://github.com/metal-toolbox/conditionorc/blob/961699928799ae5d6a1d43de442ed848b008fcc3/internal/orchestrator/updates_test.go#L147

internal/orchestrator/updates.go
joelrebel: fixed in [9616999](https://github.com/metal-toolbox/conditionorc/pull/203/commits/961699928799ae5d6a1d43de442ed848b008fcc3)

internal/orchestrator/updates.go
joelrebel: fixed in [9616999](https://github.com/metal-toolbox/conditionorc/pull/203/commits/961699928799ae5d6a1d43de442ed848b008fcc3)

internal/orchestrator/updates.go
joelrebel: We don't want to return here - since we need to identify if any status entries are missing their Condition Record.

internal/orchestrator/updates.go
joelrebel: Nice catch, I'm in agreement, although in this case since all we have is a StatusValue key and nothing more, its not possible to re-construct its equivalent Condition. I left a comment on here https://github.com/metal-toolbox/conditionorc/blob/961699928799ae5d6a1d43de442ed848b008fcc3/internal/orchestrator/updates.go#L399

pkg/api/v1/types/types.go
joelrebel: fixed in [b187837](https://github.com/metal-toolbox/conditionorc/pull/203/commits/b1878374bb35b3f03f0d26b74f7342fe652352c2)

internal/orchestrator/updates.go
DoctorVin: We can't just force a status of `failed`?

DoctorVin: As discussed we ought to send an otel event or increment some sort of metric to catch this.

DoctorVin (approved):
I think we can iterate on any remaining issues. Let's get this merged.

internal/orchestrator/updates.go
joelrebel: added span collection in [7e6d17c](https://github.com/metal-toolbox/conditionorc/pull/203/commits/7e6d17c1fab7f89e6d22b2dace24a79b54f59b5b)

PR #207, opened 2024-06-10
FS-1464: add Target field when creating a condition

Alva8756:
It looks we are missing the Target field when creating a condition.

Target remains `00000000-0000-0000-0000-000000000000` which causes [error](https://github.com/metal-toolbox/conditionorc/blob/v1.0.7/pkg/api/v1/types/types.go#L150-L152)

### DIFF
--- a/internal/store/nats.go
+++ b/internal/store/nats.go
@@ -203,6 +203,8 @@ func (n *natsStore) Create(ctx context.Context, serverID uuid.UUID, condition *r
 		state = condition.State
 	}
 
+	condition.Target = serverID
+
 	cr := ConditionRecord{
 		ID:    condition.ID,
 		State: state,


DoctorVin (changes requested):
This feels like the wrong solution to me. I would have expected the caller of this to compose its Conditions with the right contents. Fixing this here seems to be more of a band-aid.

It might be better instead to change this so we always read the server-id out of the `Condition` and return an error if we get a UUID zero-value?

Also, whatever we come up with, we should make sure we have test coverage for it.

joelrebel:
I was working on this as well and have a slightly different approach - https://github.com/metal-toolbox/conditionorc/pull/208
mind if we close this issue in favour of the one linked?

PR #208, opened 2024-06-19
TargetID, verbose logging fixes

joelrebel:
#### What does this PR do

- TargetID should be a required parameter for creating new Conditions
- Clearer errors when a condition merge fails
- Use trace instead of debug to reduce some verbosity


internal/orchestrator/notify/slack.go
```diff
@@ -44,7 +44,7 @@ func (ss *slackSender) Send(upd *v1types.ConditionUpdateEvent) error {
 		"conditionID": upd.ConditionUpdate.ConditionID.String(),
 	})
 
-	le.Debug("sending slack notification")
+	le.Trace("sending slack notification")
```
DoctorVin, line 47: Why make this change? We log at `Info` in production?

DoctorVin (commented):
I'm pretty OK with the logging change, although I'm surprised that you assessed them as overly chatty.

What is the rationale for the changes to condition constructor? I'm OK with the implications (a condition is invalid without a target, which seems obvious in hindsight). Do you believe this fixes that behavior you observed in the initial rollout of your feature? If so, can you/how did you validate your belief?

internal/orchestrator/notify/slack.go
joelrebel: Since each status update creates a log entry, this was too verbose for debugging with. We can come with some guidelines separately on what should be at debug vs trace.

joelrebel:
> What is the rationale for the changes to condition constructor? I'm OK with the implications (a condition is invalid without a target, which seems obvious in hindsight). Do you believe this fixes that behavior you observed in the initial rollout of your feature? If so, can you/how did you validate your belief?

The behaviour we saw was the first condition status update failing, with the errors below.
The merge code did what it supposed to do - validate the serverID, but the error was misleading - `no existing condition found for update`.

The fix ensures all Conditions have TargetID specified at its creation, the added test verifies the expected behaviour. As always, the changes were tested in the sandbox. As for production validation, thats internal and we can discuss that separately.


```

{
    "conditionKind": "inventory",
    "error": "Server ID's do not match: no existing condition found for update",
    "existing_state": "pending",
    "file": "handlers.go",
    "function": "UpdateCondition",
    "incoming_state": "active",
    "level": "warning",
    "line": "92",
    "msg": "condition merge failed",
    "serverID": "ede81024-f62a-4288-8730-3fab8ccea000",
    "time": "2024-06-19T13:03:00Z"
}
{
    "conditionKind": "inventory",
    "error": "ServerID in update does not match existing: condition merge error",
    "existing_state": "pending",
    "file": "handlers.go",
    "function": "UpdateCondition",
    "incoming_state": "active",
    "level": "warning",
    "line": "92",
    "msg": "condition merge failed",
    "serverID": "ede81024-f62a-4288-8730-3fab8ccea000",
    "time": "2024-06-19T13:14:26Z"
}

```

DoctorVin:
Is there a way we can add a unit test in the reconciliation pathway to catch malformed data?

joelrebel:
> Is there a way we can add a unit test in the reconciliation pathway to catch malformed data?

As discussed on slack, both the Create and Update paths make use of `MergeExisting` - https://github.com/metal-toolbox/conditionorc/blob/b202ef5e84ed7c29c547873c6c400f8dfa6c4d31/pkg/api/v1/types/types_test.go#L25
for which we have tests.

joelrebel:
A few additional fixes were required to the CI and tests to resolve various linting and test failures.

.github/workflows/push-pr-lint.yaml
```diff
@@ -5,22 +5,23 @@ jobs:
...
+        uses: actions/checkout@v4
 
-    - name: Install Go
-      uses: actions/setup-go@v4
-      with:
-        go-version-file: go.mod
+      - name: Install Go
+        uses: actions/setup-go@v4
+        with:
+          go-version-file: go.mod
```
jakeschuurmans, line 14: oh we need to do this everywhere. Many projects still have a static version to check against.

internal/orchestrator/notify/slack.go
jakeschuurmans: Guidelines for when to use debug, info, or trace are basically impossible to follow. Just have to take it case by case IMO. This is definitely a Trace IMO.

jakeschuurmans (approved):

PR #210, opened 2024-06-25
Configure chainguard to accept commits signed by GitHub

mmlb:
#### What does this PR do

Configures chainguard to accept GH's internal gpg key that is used to sign web based actions taken by the user (like commits and PR merges).


mmlb:
depends on #208 drafting until that is merged.

joelrebel (approved):

PR #217, opened 2024-07-01
Switch to testify/mock and move away from go.hollow.sh/toolbox/events

joelrebel:
#### What does this PR do

- Switches imports to `github.com/metal-toolbox/events` from `go.hollow.sh/toolbox/events`
- Changes mocks to be generated by Mockery which is in use by rivets and (imo) is nicer to work with.
- Updates all tests to use testify/mock instead of gomock.


ofaurax (approved):

pkg/api/v1/client/client_auth_test.go
```diff
@@ -38,7 +38,7 @@ func newAuthTester(t *testing.T, authToken string) (*integrationTester, finalize
 
 	ctrl := gomock.NewController(t)
```
DoctorVin, line 39: We shouldn't need this anymore?

DoctorVin (changes requested):

pkg/api/v1/client/client_auth_test.go
joelrebel: nice catch, thanks, fixed

pkg/api/v1/routes/handlers_test.go
```diff
@@ -393,6 +331,17 @@ func TestAddServer(t *testing.T) {
 			server.ServeHTTP(recorder, tc.request(t))
 
 			tc.assertResponse(t, recorder)
+
+			// Assert all expectations
```
DoctorVin, line 335: Why are we doing this here and not in other tests? The GoMock ‘finisher’ asserts the mock expectations.

DoctorVin (commented):
I don’t understand the inconsistency in asserting expectations on the mock?

DoctorVin (approved):

PR #218, opened 2024-07-02
Sets up orchestrator service routes

joelrebel:
#### What does this PR do

- Moves the oidc auth enable parameter as a top level flag for reuse
- Sets up the Orchestrator API router

Changes merge into the `fw-inband-devel` branch which will track `main` until all changes are merged into this devel branch.

### DIFF
--- a/cmd/orchestrator.go
+++ b/cmd/orchestrator.go
@@ -2,16 +2,20 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
+	"net/http"
 
 	"github.com/equinix-labs/otel-init-go/otelinit"
 
 	"github.com/metal-toolbox/conditionorc/internal/app"
+	"github.com/metal-toolbox/conditionorc/internal/fleetdb"
 	"github.com/metal-toolbox/conditionorc/internal/metrics"
 	"github.com/metal-toolbox/conditionorc/internal/model"
 	"github.com/metal-toolbox/conditionorc/internal/orchestrator"
 	"github.com/metal-toolbox/conditionorc/internal/orchestrator/notify"
+	"github.com/metal-toolbox/conditionorc/internal/server"
 	"github.com/metal-toolbox/conditionorc/internal/store"
 	"github.com/metal-toolbox/conditionorc/internal/version"
 	"github.com/metal-toolbox/rivets/events"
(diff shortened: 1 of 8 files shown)


jakeschuurmans (approved):

DoctorVin (approved):
I feel like we should have had a separate module for the orchestrator API. By avoiding this when the task is small, we've deferred work until later when the orchestrator and the  client API are larger and less-tractable to work with. 

This is OK as-is.

cmd/orchestrator.go
```diff
@@ -62,8 +66,38 @@ var cmdOrchestrator = &cobra.Command{
...
+
+		// init Orchestrator API server
+		optionsOrcAPI := []server.Option{
+			server.WithLogger(app.Logger),
+			server.WithListenAddress(app.Config.ListenAddress),
+			server.WithStore(repository),
+			server.WithFleetDBClient(fleetDBClient),
+			server.WithStreamBroker(streamBroker, app.Config.NatsOptions.PublisherSubjectPrefix),
+			server.WithConditionDefinitions(app.Config.ConditionDefinitions),
+			server.WithAsOrchestrator(facility),
```
DoctorVin, line 82: The name of the option here throws me: we're already running as an Orchestrator, and you're only specifying the facility here as an option. Why change the name of this?

joelrebel:
> I feel like we should have had a separate module for the orchestrator API. By avoiding this when the task is small, we've deferred work until later when the orchestrator and the client API are larger and less-tractable to work with.

@DoctorVin 
Its not too late to do that, and I had given it a few attempts to slice up (without breaking existing API, less code and test duplication) in different ways but was not happy with the outcome. I'll create another PR as an attempt and we can follow up there there.

internal/server/server.go
```diff
@@ -31,11 +31,13 @@ var (
 
 // Server type holds attributes of the condition orc server
 type Server struct {
-	// Logger is the app logger
+	orchestrator         bool // run in orchestrator API mode
```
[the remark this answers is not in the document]
joelrebel: would appreciate a little more elaboration

cmd/orchestrator.go
joelrebel: I'll change `WithAsOrchestrator` to `WithOrchestratorAPI` in a follow-up PR. 
I do not follow what you mean by 'Why change the name of this?' - since this is code being added.

For sake of a common convention going ahead - we currently have a `Conditions` API - which is consumed by clients requesting conditions. And going ahead, these changes introduce an `Orchestrator` API which is for consumption by controllers within our 'control plane'.

PR #221, opened 2024-07-03
[1/4] Split conditions API, add orchestrator API

joelrebel:
#### What does this PR do

For merge into the fw-inband-devel branch

- Splits up pkg/api for Conditions and Orchestrator into separate packages
- Adds Orchestrator API routes for controllers

For merge after, 
https://github.com/metal-toolbox/conditionorc/pull/222
https://github.com/metal-toolbox/rivets/pull/66 (and a tag on rivets which is to be pinned on)


joelrebel:
PR updated to limit changes to Conditions and Orchestrator API routes init.

pkg/api/v1/conditions/routes/handlers.go
```diff
@@ -366,11 +348,62 @@ func (r *Routes) firmwareInstall(c *gin.Context) (int, *v1types.ServerResponse)
...
+					rctypes.AcquireServer,
+					rctypes.PurposeFirmwareInstall,
+					"Marked for firmware install",
+				).MustMarshal(),
+				State:     rctypes.Pending,
+				CreatedAt: createTime,
+			},
+			{
+				Kind:       rctypes.FirmwareInstallInband,
+				Version:    rctypes.ConditionStructVersion,
```
DoctorVin, line 377: so we'll unconditionally queue an inband install? Will that be the default?

joelrebel: Since we insist all changes trickle into main first, this  cannot end up in main as is,
I'll comment out the additional `acquire`, `release`, `inband` conditions.

Separate changes are planned for Conditions API to compose the firmware set itself,
that would then mean the firmware set composition will dictate if inband is required or not.
Although I had not planned for introducing this change as yet, but now it seems to be more necessary than not.

internal/server/server.go
```diff
@@ -73,9 +84,10 @@ func WithListenAddress(addr string) Option {
 }
 
 // WithStreamBroker sets the event stream broker.
-func WithStreamBroker(broker events.Stream) Option {
+func WithStreamBroker(broker events.Stream, streamSubjectPrefix string) Option {
```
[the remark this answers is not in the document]
joelrebel: As of now theres no intention on changing, this could be turned into a const in rivets going ahead so we have one less configuration param being passed around.

pkg/api/v1/conditions/client/client_test.go
```diff
@@ -325,7 +332,16 @@ func TestFirmwareInstall(t *testing.T) {
 				AssetID: serverID,
 			},
 			mockStore: func(r *store.MockRepository) {
-				r.On("CreateMultiple", mock.Anything, serverID, mock.Anything, mock.Anything).
+				r.On(
+					"CreateMultiple",
+					mock.Anything,
+					serverID,
+					mock.Anything, // I haven't figured a way pass in a variable list of conditions to this r.On mock method
```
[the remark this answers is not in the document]
joelrebel: Since `CreateMultiple` accepts upto n Conditions, I was trying to get the mock `r.On` method to accept a variable list of `*rctypes.Condition` such that the test doesn't have to be updated if the Condition composition changes.

cmd/orchestrator.go
```diff
@@ -85,6 +119,13 @@ var cmdOrchestrator = &cobra.Command{
 
 		orc := orchestrator.New(options...)
 		orc.Run(ctx)
+
+		// shutdown server when orchestrator Run method returns
+		ctx, cancel := context.WithTimeout(cmd.Context(), shutdownTimeout)
```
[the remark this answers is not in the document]
joelrebel: Not really, this ctx is passed to the `*http.Server.Shutdown()` which stops accepting traffic gracefully. I don't see a way for the http.Server to shutdown based on a context cancellation.

pkg/api/v1/conditions/client/client_test.go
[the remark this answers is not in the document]
DoctorVin: I did some searching yesterday. Mockery does support variadic methods I think.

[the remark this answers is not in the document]
joelrebel: Thanks, added the fix in https://github.com/metal-toolbox/conditionorc/pull/221/commits/1484fff561d2aea85deb7a8bf439ee221422dab3

joelrebel:
@DoctorVin this is ready for merge. The point about documentation was noted, will take that up separately.

DoctorVin (dismissed):

internal/app/config.go
```diff
@@ -101,6 +101,20 @@ func (a *App) LoadConfiguration() error {
...
+	if a.v.GetBool("oidc.enabled") {
+		if len(a.Config.APIServerJWTAuth) == 0 {
+			a.Logger.Fatal("OIDC enabled without configuration")
+		}
+		a.Logger.Info("OIDC enabled")
+
+		return true
+	}
+
+	a.Logger.Info("OIDC disabled")
```
jakeschuurmans, line 114: Super nit pick.

Should OIDC being disabled be a warning?

jakeschuurmans, line 114: Super nit pick.

Should OIDC being disabled be a warning?

jakeschuurmans (changes requested):

internal/app/config.go
joelrebel: Not really, OIDC availability is deployment specific and this is informational so we can go check if its enabled or disabled from the logs.

joelrebel:
commits rebased on main

jakeschuurmans (approved):

internal/orchestrator/updates.go
```diff
@@ -616,7 +616,7 @@ func (o *Orchestrator) queueFollowingCondition(ctx context.Context, cond *rctype
 	if active != nil && active.State == rctypes.Pending {
 		byt := active.MustBytes()
 		subject := fmt.Sprintf("%s.servers.%s", o.facility, active.Kind)
-		err := o.streamBroker.Publish(ctx, subject, byt)
+		err := o.streamBroker.Publish(ctx, subject, byt, false)
```
DoctorVin, line 619: We should change the stream broker to separate Publish into 2 APIs, `Publish` and `PublishOverwrite`, each a public wrapper around a private `_publish` that takes the boolean. Otherwise you're making it harder to understand for code maintenance.

DoctorVin (changes requested):

internal/orchestrator/updates.go
DoctorVin: I did exactly this in rivets `v1.1.0`

DoctorVin (dismissed):

internal/orchestrator/updates.go
joelrebel: thanks for that, I'll pin on that version.

joelrebel:
@DoctorVin this would require another stamp

DoctorVin (approved):

PR #230, opened 2024-07-11
Reconciler fixes

joelrebel:
#### What does this PR do

- Fixes incorrect reconcile for Conditions still in queue
- Fixes incorrect cross facility reconcile

Edit:
- Final commits fixes incorrect finalized CR update attempt and StatusValue record not being purged.


internal/orchestrator/updates.go
```diff
@@ -318,10 +321,14 @@ func conditionFromUpdateEvent(evt *v1types.ConditionUpdateEvent) *rctypes.Condit
 	}
 }
 
-func filterIncompleteRecords(records []*store.ConditionRecord) []*store.ConditionRecord {
+func (o *Orchestrator) filterIncompleteRecords(records []*store.ConditionRecord) []*store.ConditionRecord {
```
DoctorVin, line 324: I'd rather not make this a member function just to avoid passing a parameter.

internal/orchestrator/updates.go
```diff
@@ -35,6 +35,9 @@ var (
 	errCompleteEvent   = errors.New("unable to complete event")
 	failedByReconciler = []byte(`{ "msg": "controller failed to process this condition in time" }`)
 	reconcilerCadence  = 1 * time.Minute
+	// Ideally set this to the JS MaxAge value.
+	// The Discard Policy should be set to Old.
+	msgMaxAgeThreshold = 24 * time.Hour
```
DoctorVin, line 40: We can read this from `nats.StreamConfig` via `nats.GetStreamInfo(...).Config.MaxAge`.

In re: discard policy, I don't know that it makes that much difference, but I'm open to the idea.

joelrebel: Yeah, this should ideally be from the stream configuration we set, as of now `MaxAge` is unlimited.
In a followup I'll look into picking it up from the configuration instead.

The Discard Policy is currently set to `New`, with `MaxAge` defined it would throw errors when accepting new messages which would not be something we like I guess - we can discuss this offline.

https://docs.nats.io/nats-concepts/jetstream/streams#discardpolicy

internal/orchestrator/updates_test.go
```diff
@@ -404,14 +406,28 @@ func TestActiveConditionsToReconcile_Updates(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// set condition CreatedAt to across the stale threshold
+	// set condition CreatedAt to across the stale threshold - condition in queue, not exceeded queue age threshold
```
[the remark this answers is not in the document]
joelrebel: Fixed borked english

internal/orchestrator/updates.go
joelrebel: thanks, fixed now

PR #232, opened 2024-07-12
WIP event handler refactor: needs tests

DoctorVin:
#### What does this PR do
Remove the vestigial `pkg/api/v1/events` code, moving (with some slight refactors) back to the orchestrator proper. I think this will make things somewhat easier for keeping the management of event updates (either from a a controller or from the reconciler go routine).

This is for awareness/feedback only. I need to write some tests to validate I didn't break anything in my work. But it currently builds, lints and tests cleanly.


internal/orchestrator/updates.go
```diff
@@ -478,25 +492,89 @@ func (o *Orchestrator) getEventsToReconcile(ctx context.Context) (evts []*v1type
...
+			o.logger.WithFields(logrus.Fields{
+				"server_id":      updEvt.ConditionUpdate.ServerID,
+				"condition_id":   updEvt.ConditionUpdate.ConditionID,
+				"condition_kind": updEvt.Kind,
+			}).Error("no existing pending/active condition found for update")
+			return nil, errors.Wrap(err, "fetching active condition to update")
+		}
+		return nil, errors.Wrap(errRetryThis, err.Error())
+	}
+
```
joelrebel, line 531: I think we're missing two cases,
- The update is applied - only if the existing CR State is non-final
- The status KV record for the event is purged, even the CR has no record of a matching ConditionID

This would cover the changes proposed in 
https://github.com/metal-toolbox/conditionorc/pull/230/commits/2ab1f3888243cfe46cde56a18288923297fdcac1#diff-b36bc375033af0c0f43e3fa155cb4078dc7dfba67b7b8986f95fb00a8cb2ce45R501

joelrebel (approved):
thanks for the clean up

PR #237, opened 2024-07-18
Vc/record event history

DoctorVin:

### DIFF
--- a/internal/orchestrator/updates.go
+++ b/internal/orchestrator/updates.go
@@ -567,10 +567,24 @@ func (o *Orchestrator) eventUpdate(ctx context.Context, evt *v1types.ConditionUp
 }
 
 func (o *Orchestrator) finalizeCondition(ctx context.Context, cond *rctypes.Condition) error {
+	// if we fail to update the event history or to delete this event from the KV,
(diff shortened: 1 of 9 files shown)


internal/orchestrator/updates.go
```diff
@@ -567,10 +567,24 @@ func (o *Orchestrator) eventUpdate(ctx context.Context, evt *v1types.ConditionUp
 }
 
 func (o *Orchestrator) finalizeCondition(ctx context.Context, cond *rctypes.Condition) error {
+	// if we fail to update the event history or to delete this event from the KV,
+	// the reconciler  will catch it later and walk this code, so return early. It
+	// is kosher to replay event history iff the contents of that history (id,
```
coffeefreak101, line 572: `iff` should be `if`?

DoctorVin: I was lazy and geeky and mean it in the "if and only if" sense. Finalizing a condition is idempotent if the non-time contents of the event are identical. If they aren't it's an error.

coffeefreak101: Ah, ok. I haven't seen that notation before.

coffeefreak101 (approved):

PR #238, opened 2024-07-19
[FS-1568]Add the draft server provision api

Alva8756:
#### What does this PR do
This is the placeholder for server provision api. Will throw an error since the API is unimplemented.
tested locally with
```
% curl -i -X POST --header 'Content-Type: application/json' localhost:9001/api/v1/serverProvision -d ""
HTTP/1.1 400 Bad Request
Content-Type: application/json; charset=utf-8
Date: Fri, 19 Jul 2024 05:54:55 GMT
Content-Length: 51

{"message":"invalid server provision payload: EOF"}% 
```
cc @iancoffey

### DIFF
--- a/pkg/api/v1/conditions/routes/handlers.go
+++ b/pkg/api/v1/conditions/routes/handlers.go
@@ -269,6 +269,26 @@ func (r *Routes) serverEnroll(c *gin.Context) (int, *v1types.ServerResponse) {
 	return st, resp
 }
 
+// @Summary Server Provision
+// @Tag Servers
+// @Description an API to perform the server provision.
+// @Accept json
+// @Produce json
+// @Success 200 {object} v1types.ServerResponse
+// @Router /serverProvision [post]
+func (r *Routes) serverProvision(c *gin.Context) (int, *v1types.ServerResponse) {
+	var sp v1types.ServerProvisionRequest
+	if err := c.ShouldBindJSON(&sp); err != nil {
+		r.logger.WithError(err).Warn("unmarshal server provision payload")
+
+		return http.StatusBadRequest, &v1types.ServerResponse{
+			Message: "invalid server provision payload: " + err.Error(),
+		}
+	}
+
+	return 501, &v1types.ServerResponse{Message: "unimplemented"}
+}
+
 // @Summary Firmware Install
 // @Tag Conditions
 // @Description Installs firmware on a device and validates with a subsequent inventory
--- a/pkg/api/v1/conditions/types/types.go
+++ b/pkg/api/v1/conditions/types/types.go
@@ -183,5 +183,4 @@ type ServerProvisionRequest struct {
 	StorageConfiguration string    `json:"StorageConfiguration,omitempty"`
 	NetworkConfiguration string    `json:"NetworkConfiguration,omitempty"`
 	Tags                 []string  `json:"Tags,omitempty"`
-	State                string    `json:"State,omitempty"`
 }
--- a/pkg/api/v1/conditions/types/types.go
+++ b/pkg/api/v1/conditions/types/types.go
@@ -170,3 +170,19 @@ func (c *ConditionUpdateEvent) MergeExisting(existing *rctypes.Condition) (*rcty
 		CreatedAt:             existing.CreatedAt,
 	}, nil
 }
+
+// ServerProvisionRequest is the request payload to provision a server.
+// Duplicated from FCP without Spec: https://github.com/equinixmetal/facility-controlplane/blob/main/pkg/api/v1/types.go#L388-L402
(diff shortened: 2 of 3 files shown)


pkg/api/v1/conditions/routes/handlers.go
```diff
@@ -269,6 +269,26 @@ func (r *Routes) serverEnroll(c *gin.Context) (int, *v1types.ServerResponse) {
...
+	var sp v1types.ServerProvisionRequest
+	if err := c.ShouldBindJSON(&sp); err != nil {
+		r.logger.WithError(err).Warn("unmarshal server provision payload")
+
+		return http.StatusBadRequest, &v1types.ServerResponse{
+			Message: "invalid server provision payload: " + err.Error(),
+		}
+	}
+
+	return 501, &v1types.ServerResponse{Message: "unimplemented"}
```
DoctorVin, line 289: I was going to quibble here, but `retryablehttp` explicitly does not retry `501` so we're all good.

pkg/api/v1/conditions/types/types.go
```diff
@@ -170,3 +170,19 @@ func (c *ConditionUpdateEvent) MergeExisting(existing *rctypes.Condition) (*rcty
...
+
+// ServerProvisionRequest is the request payload to provision a server.
+// Duplicated from FCP without Spec: https://github.com/equinixmetal/facility-controlplane/blob/main/pkg/api/v1/types.go#L388-L402
+type ServerProvisionRequest struct {
+	ServerID             uuid.UUID `json:"serverID,omitempty"`
+	InstanceID           uuid.UUID `json:"instanceID,omitempty"`
+	Hostname             string    `json:"Hostname,omitempty"`
+	ProviderType         string    `json:"ProviderType,omitempty"`
+	OperatingSystemID    string    `json:"OperatingSystemID,omitempty"`
+	InfrastructureID     []string  `json:"InfrastructureID,omitempty"`
```
DoctorVin, line 182: What's `InfrastructureID` reference?

pkg/api/v1/conditions/types/types.go
```diff
@@ -170,3 +170,19 @@ func (c *ConditionUpdateEvent) MergeExisting(existing *rctypes.Condition) (*rcty
...
+	UserData             *string   `json:"UserData,omitempty"`
+	StorageConfiguration *string   `json:"StorageConfiguration,omitempty"`
+	NetworkConfiguration *string   `json:"NetworkConfiguration,omitempty"`
```
DoctorVin, line 185: Making these pointers to string means that we're concerned about discriminating between `nil` and `"""`. What is the difference in this case?

DoctorVin (commented):
This looks fine as far as it goes. I was initially squeamish about merging it soon, but on second thought I think we're probably fine. The code only returns an error as currently (not) implemented. If no one else is opposed I say :shipit:

pkg/api/v1/conditions/types/types.go
Alva8756: Great check! I think `InfrastructureID ` duplicate with `ServerID`.
We can just keep one. I do not have strong preference. 
What do you think?

pkg/api/v1/conditions/types/types.go
Alva8756: I was follow the pattern from FCP without Spec: https://github.com/equinixmetal/facility-controlplane/blob/main/pkg/api/v1/types.go#L388-L402 to keep it consistent.

Agree with you and removed there.

pkg/api/v1/conditions/types/types.go
```diff
@@ -170,3 +170,18 @@ func (c *ConditionUpdateEvent) MergeExisting(existing *rctypes.Condition) (*rcty
 		CreatedAt:             existing.CreatedAt,
 	}, nil
 }
+
+// ServerProvisionRequest is the request payload to provision a server.
+// Duplicated from FCP without Spec: https://github.com/equinixmetal/facility-controlplane/blob/main/pkg/api/v1/types.go#L388-L402
```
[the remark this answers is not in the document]
DoctorVin: The data structure looks generic-enough to me for most things. I missed `State` and I don't think we should include it or reference it. We definitely don't want EMAPI state here and I'd argue strenuously that we don't care about any kind of state.

Regarding the openness of the code and any sort of build-time constraints that places on us, I think that's premature, and outside the scope of this particular PR. Any decision we make here will be almost immediately superseded by management, so instead let's engage with them and drive that process in a more productive forum.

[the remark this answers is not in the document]
Alva8756: Thanks for the discussion! Removed the `state` field.

DoctorVin (approved):

PR #242, opened 2024-07-29
delete unlinked status if we can't update a CR

DoctorVin:


DoctorVin:
Superseded by https://github.com/metal-toolbox/conditionorc/pull/245

PR #243, opened 2024-07-29
add an API to kick off validation

DoctorVin:
This API creates a workflow for Firmware Validation. Currently it only replicates `firmwareInstall` as there is no actual controller for firmware validation yet, but adding such should be straightforward. I also added swagger annotations and regenerated the docs.


Alva8756:
Thanks for helping on adding the server provision docs there!

Alva8756 (approved):

PR #247, opened 2024-07-30
Orchestrator API Client methods

joelrebel:
#### What does this PR do

- Updates the orchestrator/api/`conditionStatusUpdate` route handler to use `GetActiveCondition`
- Implements the Orchestrator API client methods

Pending: fix up tests

### DIFF
--- a/pkg/api/v1/orchestrator/routes/handlers.go
+++ b/pkg/api/v1/orchestrator/routes/handlers.go
@@ -296,7 +296,7 @@ func (r *Routes) taskPublish(c *gin.Context) (int, *v1types.ServerResponse) {
 	activeCond, err := r.repository.GetActiveCondition(ctx, serverID)
 	if err != nil {
 		if errors.Is(err, store.ErrConditionNotFound) {
-			return http.StatusBadRequest, &v1types.ServerResponse{
+			return http.StatusNotFound, &v1types.ServerResponse{
 				Message: err.Error(),
 			}
 		}
--- a/pkg/api/v1/orchestrator/routes/handlers_test.go
+++ b/pkg/api/v1/orchestrator/routes/handlers_test.go
@@ -578,7 +578,7 @@ func TestTaskPublish(t *testing.T) {
 				return request
 			},
 			assertResponse: func(t *testing.T, r *httptest.ResponseRecorder) {
-				assert.Equal(t, http.StatusBadRequest, r.Code)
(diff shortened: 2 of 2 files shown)


pkg/api/v1/orchestrator/routes/handlers.go
```diff
@@ -92,34 +92,28 @@ func (r *Routes) conditionStatusUpdate(c *gin.Context) (int, *v1types.ServerResp
 	}
 
 	// expect condition to be in an incomplete state to accept this update.
-	cr, err := r.repository.Get(ctx, serverID)
+	cond, err := r.repository.GetActiveCondition(ctx, serverID)
 	if err != nil {
-		return http.StatusInternalServerError, &v1types.ServerResponse{
-			Message: "condition lookup: " + err.Error(),
+		if errors.Is(err, store.ErrConditionNotFound) {
+			return http.StatusBadRequest, &v1types.ServerResponse{
```
DoctorVin, line 98: Arguably should be `StatusNotFound`.

pkg/api/v1/orchestrator/routes/handlers.go
```diff
@@ -92,34 +92,28 @@ func (r *Routes) conditionStatusUpdate(c *gin.Context) (int, *v1types.ServerResp
...
-			if !rctypes.StateIsComplete(cond.State) {
-				updateable = true
-			}
+		return http.StatusInternalServerError, &v1types.ServerResponse{
+			Message: "condition lookup: " + err.Error(),
 		}
 	}
 
-	if !found {
+	if rctypes.StateIsComplete(cond.State) {
```
DoctorVin, line 108: Prior to this, in light of our discoveries yesterday, maybe we should check that the update is actually applicable to the active condition's ID? That is `cond.ID == conditionID`?

DoctorVin (commented):
I think the return code is an elective, but I would like the protection of checking the incoming condition id against active condition id, just in case.

pkg/api/v1/orchestrator/routes/handlers.go
joelrebel: thanks, I was conflicted between it being a incorrect request vs a 404. Updated all the handlers to return 404s.

pkg/api/v1/orchestrator/routes/handlers.go
joelrebel: added

DoctorVin (approved):

PR #265, opened 2024-09-02
api/conditions/handlers: Condition to include TraceID, SpanID, Client ID attributes

joelrebel:
#### What does this PR do

This enables tracing and linking Conditions executed by Controllers and
managed by the Orchestrator. So far, Span, TraceIDs were included in the
the NATS message header when a Condition gets published to the
Jetstream. 

Since not all Conditions are published onto the Jetstream (firmwareInstallInband) and 
we'd like to ensure the traces are linked. The TraceID, SpanID is included in the Condition payload itself and the NATS message  header can be deprecated.

 The Client ID information is included to ensure we can identify the
 client that queued the request.

Screenshot 2024-09-02 at 16 32 48

NOTE: Pin on rivets tag before merge.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/google/uuid v1.6.0
 	github.com/hashicorp/go-retryablehttp v0.7.7
 	github.com/metal-toolbox/fleetdb v1.19.3
-	github.com/metal-toolbox/rivets v1.3.3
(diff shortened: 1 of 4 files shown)


ofaurax (dismissed):

ofaurax (approved):

joelrebel:
rebased on main

ofaurax (approved):

PR #266, opened 2024-09-04
api/conditions: fixes incorrect firmware composition for oob installs

joelrebel:
#### What does this PR do
The CONDITION_API_FEATURE_INBAND_FIRMWARE env variable when enabled would result in an empty list of Conditions if all the firmware in the set are to be installed out of band. This fixes the issue and adds tests.

### DIFF
--- a/pkg/api/v1/conditions/routes/handlers.go
+++ b/pkg/api/v1/conditions/routes/handlers.go
@@ -480,8 +480,10 @@ func (r *Routes) firmwareInstallComposite(
 		Conditions: []*rctypes.Condition{},
 	}
 
-	// under feature flag until this is confirmed to be working as expected
-	// - if set - include require inband - if firmwares shares that requirement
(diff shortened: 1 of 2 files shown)


pkg/api/v1/conditions/routes/handlers.go
```diff
@@ -481,21 +481,18 @@ func (r *Routes) firmwareInstallComposite(
 	}
 
 	// under feature flag until this is confirmed to be working as expected
+	// - if set - include require inband - if firmwares shares that requirement
```
ofaurax, line 484: I don't understand this comment

ofaurax (approved):
Minor comment, but overall ok

pkg/api/v1/conditions/routes/handlers.go
joelrebel: thanks, I've now updated the comment

PR #303, opened 2025-02-24
Add some additional error messaging around biosControl

mattcburns:
#### What does this PR do

This PR adds the failed request body to the error message when there is a failure to parse the request body sent to biosControl. This is necessary because the current error message does not provide what the incorrect body looks like, just that it wasn't able to parse it. When working with requests that may have been mangled by an intermediate service, it helps to have this as part of the error response for debugging purposes.

### DIFF
--- a/.github/workflows/push-pr-lint.yaml
+++ b/.github/workflows/push-pr-lint.yaml
@@ -17,7 +17,7 @@ jobs:
         uses: golangci/golangci-lint-action@v6
         with:
           args: --config .golangci.yml --timeout 2m
-          version: v1.61.0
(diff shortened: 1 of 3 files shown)


coffeefreak101 (dismissed):
It might be useful or helpful to also add the body to the "unmarshal biosControl payload" log message above.

mattcburns:
> It might be useful or helpful to also add the body to the "unmarshal biosControl payload" log message above.

Great idea! I went ahead and added the body to the logging error message too.

coffeefreak101 (approved):

